PR #2, opened 2018-07-17
audit support (very wip)

little-dude:
still working on message parsing for now. not much to see.


little-dude:
Still working on that. Here is an update if anyone is interested.

- I have to code to parse audit messages
- I've been reworking the whole error handling to use failure, so that when message parsing fails, we have a chain of errors that tells exactly where the parser failed (malformed NLA? truncated message? invalid field value?)

As I was playing with audit I realized I forgot a major feature: listening on multicast netlink sockets. Without that, we cannot to things like `ip monitor` or listening for audit event. That's what I'm working on now.

PR #7, opened 2018-11-24
Implement address support

mhristache:
This commit is adding a higher level API to `iproute2` that can be used for the same purpose as `ip addr` util.

This commit is also fixing a bug in the `rtnetlink` lib.

Note: I've seen some strange behavior when sharing a handle between link and address (it was randomly failing to run the second future) so there is probably some bug in the connection handling part. The workaround was to create two handles (see the `flush_addresses` example).

### DIFF
--- a/iproute2/src/addr/mod.rs
+++ b/iproute2/src/addr/mod.rs
@@ -1,0 +1,248 @@
+use errors::NetlinkIpError;
+use failure::Error;
+use ipnetwork::IpNetwork;
+use rtnetlink::{AddressCacheInfo, AddressHeader, AddressMessage, AddressNla};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+
+mod handle;
+pub use self::handle::*;
+
+mod add;
+pub use self::add::*;
+
+mod del;
+pub use self::del::*;
+
+mod get;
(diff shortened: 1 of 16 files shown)


little-dude:
Awesome, thanks for the PR! I'm out for the weekend, so I'll review / merge on Monday if you don't mind. Thanks again!

mhristache:
Sure, no problem. Thanks!

iproute2/src/addr/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+        if let Some(ip) = self.ifa_local {
+            Ok(IpNetwork::new(ip, self.prefix_len())?)
+        } else if let Some(ip) = self.ifa_address {
+            Ok(IpNetwork::new(ip, self.prefix_len())?)
+        } else {
+            bail!("IP address not present")
+        }
+    }
+
+    pub fn from_address_message(value: AddressMessage) -> Result<Self, NetlinkIpError> {
```
little-dude: Any reason for not making this a `From<AddressMessage>` impl instead?

little-dude: Oh, it would require `TryFrom` I see.
We should use that when it stabilizes.

PR #9, opened 2018-11-27
Ignore unexpected bytes in Nla buffers

### LINKED ISSUE #3: MalformedNlaValue error
Hi,

Thanks for the great work!

I've received the `MalformedNlaValue` on the example case 
```
handle.link().get().execute().and_then(|links| {
            println!("{:#?}", links);
            Ok(())
 });
```

The reason is that the size doesn't match in `NativeNla` `from_bytes` check [here](https://github.com/little-dude/netlink/blob/master/rtnetlink/src/packets/nla.rs#L257)
I was getting this error on `LinkStats<u32>` (92 bytes vs. required 96 bytes). 

It seems like, structs may have the different size depending on the Linux version (?).

### LINKED ISSUE #8: Does not work with new kernels
The newer version of linux kernel [have added support for direct broadcast forwarding](https://patchwork.ozlabs.org/patch/950046/) and that is causing the rtnetlink parsing of `LinkAfInetNla` to fail:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: MalformedNlaValue', libcore/result.rs:1009:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

The `LinkInetDevConf` struct is basically missing the newly added field (`IPV4_DEVCONF_BC_FORWARDING`) and so it cannot be created from the received NLA

```rust
impl<'buffer, T: AsRef<[u8]> + ?Sized> Parseable<LinkAfInetNla> for NlaBuffer<&'buffer T> {
    fn parse(&self) -> Result<LinkAfInetNla> {
        use self::LinkAfInetNla::*;
        let payload = self.value();
        Ok(match self.kind() {
            IFLA_INET_UNSPEC => Unspec(payload.to_vec()),
            IFLA_INET_CONF => DevConf(LinkInetDevConf::from_bytes(payload)?),
            _ => Other(<Self as Parseable<DefaultNla>>::parse(self)?),
        })
    }
}

pub(crate) trait NativeNla
where
    Self: Copy,
{
    fn from_bytes(buf: &[u8]) -> Result<Self> {
        if buf.len() != size_of::<Self>() {
            return Err(Error::MalformedNlaValue);
        }
        Ok(unsafe { ptr::read(buf.as_ptr() as *const Self) })
    }
```
This is how an strace of the NLA looks like:

```
{{nla_len=136, nla_type=AF_INET}, {{nla_len=132, nla_type=IF    LA_INET_CONF}, [[IPV4_DEVCONF_FORWARDING-1] = 1, [IPV4_DEVCONF_MC_FORWARDING-1] = 0, [IPV4_DEVCONF_PROXY_ARP-1] = 0, [IPV4_DEVCONF_ACCEPT_REDIRECTS-1] = 1, [IPV4_DEVCONF_SECURE_REDIRECTS-1] = 1, [IPV4_DEVCON    F_SEND_REDIRECTS-1] = 1, [IPV4_DEVCONF_SHARED_MEDIA-1] = 1, [IPV4_DEVCONF_RP_FILTER-1] = 0, [IPV4_DEVCONF_ACCEPT_SOURCE_ROUTE-1] = 1, [IPV4_DEVCONF_BOOTP_RELAY-1] = 0, [IPV4_DEVCONF_LOG_MARTIANS-1] = 0, [IPV    4_DEVCONF_TAG-1] = 0, [IPV4_DEVCONF_ARPFILTER-1] = 0, [IPV4_DEVCONF_MEDIUM_ID-1] = 0, [IPV4_DEVCONF_NOXFRM-1] = 0, [IPV4_DEVCONF_NOPOLICY-1] = 0, [IPV4_DEVCONF_FORCE_IGMP_VERSION-1] = 0, [IPV4_DEVCONF_ARP_AN    NOUNCE-1] = 0, [IPV4_DEVCONF_ARP_IGNORE-1] = 0, [IPV4_DEVCONF_PROMOTE_SECONDARIES-1] = 0, [IPV4_DEVCONF_ARP_ACCEPT-1] = 0, [IPV4_DEVCONF_ARP_NOTIFY-1] = 0, [IPV4_DEVCONF_ACCEPT_LOCAL-1] = 0, [IPV4_DEVCONF_SR    C_VMARK-1] = 0, [IPV4_DEVCONF_PROXY_ARP_PVLAN-1] = 0, [IPV4_DEVCONF_ROUTE_LOCALNET-1] = 0, [IPV4_DEVCONF_IGMPV2_UNSOLICITED_REPORT_INTERVAL-1] = 10000, [IPV4_DEVCONF_IGMPV3_UNSOLICITED_REPORT_INTERVAL-1] = 1    000, [IPV4_DEVCONF_IGNORE_ROUTES_WITH_LINKDOWN-1] = 0, [IPV4_DEVCONF_DROP_UNICAST_IN_L2_MULTICAST-1] = 0, [IPV4_DEVCONF_DROP_GRATUITOUS_ARP-1] = 0, [IPV4_DEVCONF_BC_FORWARDING-1] = 0]}}
```

I guess we need a bit more fancy parsing logic here.

I see the same parsing approach is used all over the code so I think these kind of issues will continuously pop up if the parsing logic is not improved to accept unknown fields.

mhristache:
This commit is changing the logic of the NativeNla::from_bytes()
method to truncate or pad with 0 the input buffer to match the size of
the result struct.

The result is that fields which are not defined in the struct are ignored
and fields which are defined in the struct but missing in the buffer are
set to 0. In this way the parsing cannot fail.

Note: I think this is a quite safe way to tackle the issue but I am not 100% sure about it. So looking forward for your opinion.

Fix #3
Fix #8

### DIFF
--- a/rtnetlink/src/packets/nla.rs
+++ b/rtnetlink/src/packets/nla.rs
@@ -254,10 +254,17 @@ where
     Self: Copy,
 {
     fn from_bytes(buf: &[u8]) -> Result<Self> {
-        if buf.len() != size_of::<Self>() {
-            return Err(Error::MalformedNlaValue);
+        // some expected fields might not present in the buffer (issue #3)
+        // or some additional/unexpected fields might be present in the buffer (issue #8)
+        // to avoid parsing errors the input buffer is truncated or padded with 0
+        // to match the size of Self so that parsing cannot fail
+        let mut my_buf = vec![0u8; size_of::<Self>()];
+        if buf.len() > size_of::<Self>() {
+            my_buf.copy_from_slice(&buf[..size_of::<Self>()]);
+        } else {
+            my_buf[..buf.len()].copy_from_slice(buf);
         }
-        Ok(unsafe { ptr::read(buf.as_ptr() as *const Self) })
+        Ok(unsafe { ptr::read(my_buf.as_ptr() as *const Self) })
     }
 
     fn to_bytes(&self, buf: &mut [u8]) {


mhristache:
ping @little-dude

little-dude:
Sorry! I'm a bit hesitant regarding this: I understand that as a user, it's annoying to have a parsing error for a message due to an obscure NLA that we might not even care about. But panicking also helps catching bugs. I'll merge this, and maybe add a compilation flag  later, to enable the panic behavior.

Thanks!

PR #11, opened 2018-12-05
Implement best-effort parsing of NL attributes

jvff:
This is a simple attempt at making sure parsing succeeds even though some attributes fail. Any attributes that fail to parse are logged as warnings instead.

I'm not sure if there's a way to reduce the repeated code introduced in both `link/message.rs` and `link/nlas/link_infos.rs`. If there are any suggestions I can try to implement them.

I also added a `FIXME` because it might be interesting in the future to make any errors available for inspection by the user.


little-dude:
Sorry for being late on this... Thank you very much this is something I wanted to do actually :)

PR #12, opened 2018-12-13
WIP: TC support

MarkPash:
Trying to add TC support to rtnetlink and could use some guidance.
Most of this is adapted from the rest of the code, still new to rust so I don't know best practices/idiomatic code.
So far I have only tested with qdisc messages, and they mostly work. There are nested NLAs that depend on the value of the parent NLA and I don't know how best to approach that.

PR #15, opened 2018-12-28
TC support

little-dude:
A rebased version of https://github.com/little-dude/netlink/pull/12

@MarkPash I can give you write access to this branch, or you can use it in #12, let me know what you prefer.


MarkPash:
Oh this is excellent thank you, I'll take the write access please :)

little-dude:
Done. You should have received an invite. I'll close #12 then.

netlink-packet/src/rtnl/tc/message.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+}
+
+impl Emitable for TcHeader {
+    fn buffer_len(&self) -> usize {
+        TC_HEADER_LEN
+    }
+
+    fn emit(&self, buffer: &mut [u8]) {
```
little-dude: Should we set the padding fields to `0` here? In theory they can be ignored but I don't know, it feels weird to have uninitialized bytes in the buffer.

netlink-packet/src/rtnl/tc/nla.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub backlog: u32,
+    /// number of dropped packets
+    pub drops: u32,
+    /// number of requeues
+    pub requeues: u32,
+    /// number of enqueues over the limit
+    pub overlimits: u32,
+}
+
+impl NativeNla for TcStatsQueue {}
```
little-dude: There are [a few other strucs](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/gen_stats.h#L39) that may be worth adding now.

little-dude:
While reviewing I made a few minor changes, I hope you don't mind

little-dude:
Answering your question in the other thread:

> `Options` has to be parsed differently based on `Kind`, as the kind is the Queue Discipline and each kind of qdisc has its own unique set of options/tuning knobs, so this is what I need help with figuring out.
The same goes for `StatsApp` as it's the statistics that are unique to that qdisc.

This sounds similar to the link infos attributes but I'm not 100% sure. Are the `Options` and `Kind` attribute _nested_? Or does the `Options` attribute _follow_ the kind attribute.

In the case of the link info, we have two attributes that always go by pair: `IFLA_INFO_KIND` followed by `IFLA_INFO_DATA`, and to parse the info data, we first parse the  info kind. See [here](https://github.com/little-dude/netlink/blob/8de810c7ab8226165d2fe4d89383d421e39c4bd0/netlink-packet/src/rtnl/link/nlas/link_infos.rs#L97-L174). The consequence is that we cannot impl `Parseable<LinkInfo>` for `NlaBuffer`, and instead we impl `Parseable<Vec<LinkInfo>>` for `NlaBuffer`. It turns out that this is not too much of a limitation, because you _usually_ want to parse all the NLAs, not just one.

Does that help? Or is this case slightly different?

MarkPash:
> While reviewing I made a few minor changes, I hope you don't mind

Not at all, please review the code and suggest/make any changes you like

> In the case of the link info, we have two attributes that always go by pair: `IFLA_INFO_KIND` followed by `IFLA_INFO_DATA`, and to parse the info data, we first parse the info kind. See here. The consequence is that we cannot impl `Parseable<LinkInfo>` for `NlaBuffer`, and instead we impl `Parseable<Vec<LinkInfo>>` for `NlaBuffer`. It turns out that this is not too much of a limitation, because you usually want to parse all the NLAs, not just one.

> Does that help? Or is this case slightly different?

Yes it did! So I now locally have something that will parse the `Options` matching on the `Kind` with a `TcOptionsNla` enum containing a couple of variants that match what I have running on my system. I'm still having it output `Vec<u8>` but at least the dependency issue is solved, I now just need to make the structs for each kind of options nla.

The code is getting a little messy so I'm restructuring and I'll push when I have something that can parse one or 2 of the option types. I assume I can keep committing my progress and we can squash the commits later?

little-dude:
> I now locally have something that will parse the Options matching on the Kind with a TcOptionsNla enum containing a couple of variants that match what I have running on my system. I'm still having it output Vec<u8> but at least the dependency issue is solved, I now just need to make the structs for each kind of options nla.

Nice!

> I assume I can keep committing my progress and we can squash the commits later?

Feel free to rework the history as you see fit yes, I don't follow any specific rule wrt to this.

-------------

I forgot to mention that for debugging netlink messages, I also use `nlmon` a lot. You may already be aware of it, but just in case, here is how you can set that up:

```
ip link add  nlmon0 type nlmon
ip link set dev nlmon0 up
tcpdump -i nlmon0 -w netlink.pcap # capture your packets
```

I found wireshark dissectors pretty useful sometimes.

little-dude:
Also, I remove `NativeNla` in https://github.com/little-dude/netlink/pull/17, because I wasn't sure whether it was really safe and clippy flagged it as an error. That means you'll have to write some pretty tedious code unfortunately :( If there are attribute you don't really care about feel free to not parse them though, and keep them as `Vec<u8>` for now. We can always revisit later.

little-dude:
rebased and removed the `NativeNla` uses.

little-dude:
I'm going to merge this as is, but feel free to add what's missing in another PR. Thanks a lot for your work!

PR #18, opened 2018-12-29
New release

little-dude:
Now that there's some support for audit, and that the code has been refactored, I'd like to make a new release.

Todo:

- [ ] more documentation
- [X] rust 2018
- [ ] make sure all the examples are still working
- [ ] more examples for audit
- [x] update the cargo manifests
- [ ] add changelog
- [ ] update the readme


little-dude:
There's too much to do before we can make a release. I'll merge the changes incrementally instead.

mhristache:
> 
> @mhristache what do you think of 670baa9 ?
> I'd like to not have public dependencies on ipnetwork or eui48, because:
> 
>     I'd like to avoid having too many dependencies in general
>     I'd like to avoid public dependencies in particular
>     they are not that widely used, and I don't want users to have to include them in their project
>     ipnetwork is advertised as "work in progress"
> 
> But on the other hand, that makes the API a bit uglier, so I don't know. What are your thought?

I am ok with the changes.

P.S.: Sorry for my late reply. I was on (a long) vacation.

PR #51, opened 2019-09-24
Remove dependence on kernel structs

pinkisemils:
We've been having issues with running this library on multiple different kernels when receiving netlink messages which layout is kernel specific - if the length of a struct is different on a different kernel, then the serialization will fail and panic. The changes I've got here just change the the different NLA enumerations to not actually contain any kernel specific structures. Unfortunately, this means that all of the different structs that may be different on different kernels are now represented as byte vectors, but the user of the library may still use the existing code to parse the byte vectors into the already defined structs. 
If these changes are too intrusive, I'm open to finding a better solution :)

---
This change is [Reviewable](https://reviewable.io/reviews/little-dude/netlink/51)

### DIFF
--- a/netlink-packet-route/src/rtnl/link/nlas/inet6/mod.rs
+++ b/netlink-packet-route/src/rtnl/link/nlas/inet6/mod.rs
@@ -8,7 +8,7 @@ use crate::{
             IFLA_INET6_ICMP6STATS, IFLA_INET6_STATS, IFLA_INET6_TOKEN, IFLA_INET6_UNSPEC,
         },
         nla::{DefaultNla, Nla, NlaBuffer},
-        traits::{Emitable, Parseable},
+        traits::Parseable,
         utils::{parse_ipv6, parse_u32, parse_u8},
     },
     DecodeError,
@@ -26,14 +26,14 @@ pub use self::stats::*;
 #[derive(Clone, Eq, PartialEq, Debug)]
 pub enum LinkAfInet6Nla {
     Flags(u32),
-    CacheInfo(LinkInet6CacheInfo),
+    CacheInfo(Vec<u8>),
     // LinkInet6DevConf is big (198 bytes), so we're wasting a space for each variant without a box.
-    DevConf(Box<LinkInet6DevConf>),
+    DevConf(Vec<u8>),
     Unspec(Vec<u8>),
     // LinkInet6Stats is huge (288 bytes), so we're wasting a *lot* of space for each variant without a
     // box.
-    Stats(Box<LinkInet6Stats>),
-    IcmpStats(LinkIcmp6Stats),
+    Stats(Vec<u8>),
+    IcmpStats(Vec<u8>),
     Token([u8; 16]),
     AddrGenMode(u8),
     Other(DefaultNla),
@@ -44,10 +44,10 @@ impl Nla for LinkAfInet6Nla {
         use self::LinkAfInet6Nla::*;
         match *self {
             Unspec(ref bytes) => bytes.len(),
-            CacheInfo(ref cache_info) => cache_info.buffer_len(),
-            DevConf(ref dev_conf) => dev_conf.buffer_len(),
-            Stats(ref stats) => stats.buffer_len(),
-            IcmpStats(ref icmp_stats) => icmp_stats.buffer_len(),
+            CacheInfo(ref cache_info) => cache_info.len(),
+            DevConf(ref dev_conf) => dev_conf.len(),
+            Stats(ref stats) => stats.len(),
+            IcmpStats(ref icmp_stats) => icmp_stats.len(),
             Flags(_) => 4,
             Token(_) => 16,
             AddrGenMode(_) => 1,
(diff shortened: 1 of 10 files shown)


little-dude:
Hi @pinkisemils, thanks for the PR. I'll review it tonight.

> If these changes are too intrusive, I'm open to finding a better solution

Well I understand the pain you're having so I think I'd be in favor of these changes. I'll comment more in details later.

netlink-packet-route/src/rtnl/link/nlas/inet6/mod.rs
```diff
@@ -26,14 +26,14 @@ pub use self::stats::*;
 #[derive(Clone, Eq, PartialEq, Debug)]
 pub enum LinkAfInet6Nla {
     Flags(u32),
-    CacheInfo(LinkInet6CacheInfo),
+    CacheInfo(Vec<u8>),
     // LinkInet6DevConf is big (198 bytes), so we're wasting a space for each variant without a box.
```
little-dude: I think this comment should be removed

netlink-packet-route/src/rtnl/link/nlas/inet6/mod.rs
```diff
@@ -26,14 +26,14 @@ pub use self::stats::*;
...
-    DevConf(Box<LinkInet6DevConf>),
+    DevConf(Vec<u8>),
     Unspec(Vec<u8>),
     // LinkInet6Stats is huge (288 bytes), so we're wasting a *lot* of space for each variant without a
     // box.
```
little-dude: This comment should be removed as well

little-dude:
Also I think I might re-introduce these in the future but behind a feature flag. Not sure yet.

little-dude:
> We've been having issues with running this library on multiple different kernels

If I may ask are you using only the `netlink-packet-route` crate with `netlink-sys` or are you using the `rtnetlink` crate which provides integration with Tokio? Also if you have feedback regarding those crates I'm curious. I wrote them with the idea of using them for a project but netlink turned out to be an entire project by itself, so I never to really use these crates yet.

pinkisemils:
>If I may ask are you using only the netlink-packet-route crate with netlink-sys or are you using the rtnetlink crate which provides integration with Tokio? Also if you have feedback regarding those crates I'm curious. I wrote them with the idea of using them for a project but netlink turned out to be an entire project by itself, so I never to really use these crates yet.

We currently use and older forked version with the changes I've got here, but to be blunt about it, we use `rtnetlink`, `netlink-proto`, `netlinkt-packet` and `netlink-sys`. You can have a look yourself [here](https://github.com/mullvad/mullvadvpn-app/tree/master/talpid-core/Cargo.toml). We use your netlink libraries to [monitor changes to the routing table](https://github.com/mullvad/mullvadvpn-app/blob/master/talpid-core/src/routing/linux/change_listener.rs#L40) and monitoring if there are any interfaces up to establish [whether a host is offline or online](https://github.com/mullvad/mullvadvpn-app/blob/master/talpid-core/src/offline/linux.rs).  Eventually, we'll want to add changes to add and change routes through your library as well, but for now these plans are very much _pie in the sky_. As far as usability of the library, I find it to be better than most tokio driven libraries. I am personally of the opinion that a lot of async crates could do a better job of being less tightly coupled with tokio, and yours is doing a good job at that. As far as the code quality is concerned, it's easy to work on. I would say that a thing that could be improved would be to return errors instead of panicking when encountering unexpected data. Since we are developing a daemon that is supposed to start on boot and never shut down, any non-recoverable is a show-stopper. I can't say I've got any specific or pragmatic advice besides that.

Besides that, I've force pushed the changes with the old comments removed. There seems to be an issue in the CI that `lazy_static` is not declared as a dev dependency, hence the `netlink-packet-audit` job is failing. Should I add a commit to fix that here, or is this something that will be fixed in the newer releases of Rust?

little-dude:
I'm very happy that these crates are useful to someone :) 

> I am personally of the opinion that a lot of async crates could do a better job of being less tightly coupled with tokio

[You're not the only one](https://github.com/heim-rs/heim/issues/75). Right now, we're pretty tightly coupled to `tokio` because `netlink-sys` uses [`PollEvented`](https://github.com/tokio-rs/tokio/blob/d1f60ac4c69792211f9c9f3bcff1559bc0cb837a/tokio-net/src/util/poll_evented.rs#L105). It's a little unclear to make my crates compatible with other runtimes, so I'm waiting for the dust to settle in the async ecosystem.

> As far as the code quality is concerned, it's easy to work on. I would say that a thing that could be improved would be to return errors instead of panicking when encountering unexpected data.

I don't know how old the fork your using is, but currently de-serialization should not panic anymore upon encountering unexpected data. If you have examples of input that trigger a panic, could you file a bug please?

> There seems to be an issue in the CI that lazy_static is not declared as a dev dependency, hence the netlink-packet-audit job is failing. Should I add a commit to fix that here, or is this something that will be fixed in the newer releases of Rust?

That's fine I'll fix that later.

PR #65, opened 2020-02-06
Add starter packet for handling fib rules

hermantolim:
Minimal solution for #5. Examples included

### DIFF
--- a/netlink-packet-route/src/rtnl/constants.rs
+++ b/netlink-packet-route/src/rtnl/constants.rs
@@ -646,3 +646,54 @@ pub const NDA_IFINDEX: u16 = 8;
 pub const NDA_MASTER: u16 = 9;
 pub const NDA_LINK_NETNSID: u16 = 10;
 pub const NDA_SRC_VNI: u16 = 11;
+
+/// see `https://github.com/torvalds/linux/blob/master/include/uapi/linux/fib_rules.h`
+
+pub const FR_ACT_UNSPEC: u8 = 0;
+pub const FR_ACT_TO_TBL: u8 = 1; /* Pass to fixed table */
+pub const FR_ACT_GOTO: u8 = 2; /* Jump to another rule */
+pub const FR_ACT_NOP: u8 = 3; /* No operation */
+pub const FR_ACT_RES3: u8 = 4;
+pub const FR_ACT_RES4: u8 = 5;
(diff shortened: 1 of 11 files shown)


little-dude:
Oh wow thank you so much @hermantolim ! I had a quick look and it looks pretty good. But I'm a little overwhelmed with work right now, so I'll review in more depth first thing this week end.

jethrogb:
According to man 7 rtnetlink these messages just carry a struct rtmsg, why aren't you reusing the `RouteMessage` type?

jethrogb:
I guess the NLAs are completely different for routing rules, so this does make sense.

hermantolim:
@jethrogb yes, actually we could actually re-use the `RouteMessageBuffer` to parse the rule, by also implementing 

```rust
impl<'a, T: AsRef<[u8]> + 'a> Parseable<RouteMessageBuffer<&'a T>> for Vec<rtnl::rule::nlas::Nla>
```

to parse NLAS. but for the header event though the structure are the same, the flag values are different.

I decide to replicate the struct from `rtnl::route` (mostly renaming and remapping the constants), to follow the code structure. since we are mapping tons of constant values here

hermantolim:
also just to be save, when checked with strace, the message send always different in headers struct

```sh
strace -yy -e trace=socket,sendto ip rule show table 0
```
and
```sh
strace -yy -e trace=socket,sendto ip route show table 0
```

netlink-packet-route/src/rtnl/constants.rs
```diff
@@ -646,3 +646,54 @@ pub const NDA_IFINDEX: u16 = 8;
 pub const NDA_MASTER: u16 = 9;
 pub const NDA_LINK_NETNSID: u16 = 10;
 pub const NDA_SRC_VNI: u16 = 11;
+
+/// see `https://github.com/torvalds/linux/blob/master/include/uapi/linux/fib_rules.h`
+
+pub const FR_ACT_UNSPEC: u8 = 0;
+pub const FR_ACT_TO_TBL: u8 = 1; /* Pass to fixed table */
```
little-dude: Would you mind turning the comments into doc with `///` ?

little-dude:
Merged, thanks again for working on this!

little-dude:
Just published a new version of `netlink-packet-route` with your changes: https://crates.io/crates/netlink-packet-route

PR #69, opened 2020-02-27
Allow ErrorMessage Conversion into std::io::Error

hermantolim:
according to [netlink(7)](https://linux.die.net/man/7/netlink), the **`NLMSG_ERROR`** return sys errno code in **`Negative`** value. This can be converted into **`std::io::Error`** using **`from_raw_os_error`** and the absolute value of the errno code for a better error handling since currently the error message is displayed in raw value.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 **/target
 **/*.rs.bk
 Cargo.lock
+.idea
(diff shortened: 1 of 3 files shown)


little-dude:
Very nice, this will help for debugging, thanks!

PR #79, opened 2020-05-28
Allowing dumping link with filter mask

cathay4t:
The linux bridge vlan filter information is only available when
`RTEXT_FILTER_BRVLAN` set in `Nla::ExtMask`.

This patch introduce `LinkGetRequest::set_filter_mask()` method allowing
such usage via:

    let (connection, handle, _) = rtnetlink::new_connection().unwrap();
    tokio::spawn(connection);
    let mut links = handle.link().get().set_filter_mask(RTEXT_FILTER_BRVLAN)
        .execute();

The similar libnetlink function is `rtnl_linkdump_req_filter()`.

Also changed these constants to u32 in compliance of libnetlink and
kernel codes:

 * `RTEXT_FILTER_VF`
 * `RTEXT_FILTER_BRVLAN`
 * `RTEXT_FILTER_BRVLAN_COMPRESSED`
 * `RTEXT_FILTER_SKIP_STATS`

Example code for query linux bridge vlan information is included in
`rtnetlink/examples/get_links.rs` as `dump_bridge_filter_info()`.

### DIFF
--- a/rtnetlink/Cargo.toml
+++ b/rtnetlink/Cargo.toml
@@ -22,3 +22,4 @@ netlink-proto = { path = "../netlink-proto", version = "0.2" }
 env_logger = "0.7.1"
 ipnetwork = "0.15.1"
 tokio = { version = "0.2.6", features = ["macros", "rt-core"] }
+netlink-sys = {path = "../netlink-sys", version = "0.2" }
--- a/rtnetlink/src/link/get.rs
+++ b/rtnetlink/src/link/get.rs
@@ -6,8 +6,8 @@ use futures::{
 
 use crate::{
     packet::{
-        nlas::link::Nla, LinkMessage, NetlinkMessage, NetlinkPayload, RtnlMessage, NLM_F_DUMP,
-        NLM_F_REQUEST,
+        nlas::link::Nla, LinkHeader, LinkMessage, NetlinkMessage, NetlinkPayload, RtnlMessage,
+        NLM_F_DUMP, NLM_F_REQUEST,
     },
     Error, Handle,
 };
@@ -32,6 +32,13 @@ impl LinkGetRequest {
             dump: true,
             filter_builder: LinkFilterBuilder::new(),
         }
+    }
+
+    /// Setting filter mask(e.g. RTEXT_FILTER_BRVLAN and etc)
+    pub fn set_filter_mask(mut self, family: u8, filter_mask: u32) -> Self {
+        self.message.header.interface_family = family;
+        self.message.nlas.push(Nla::ExtMask(filter_mask));
+        self
     }
 
     /// Execute the request
--- a/netlink-sys/src/constants.rs
+++ b/netlink-sys/src/constants.rs
@@ -511,9 +511,9 @@ pub const RTAX_FEATURE_MASK: int = 15;
 #[allow(overflowing_literals)]
 pub const TCM_IFINDEX_MAGIC_BLOCK: int = 0xffff_ffff;
 pub const TCA_FLAG_LARGE_DUMP_ON: int = 1;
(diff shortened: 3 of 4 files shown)


rtnetlink/src/link/get.rs
```diff
@@ -34,6 +34,17 @@ impl LinkGetRequest {
         }
     }
 
+    /// Setting filter mask(e.g. RTEXT_FILTER_BRVLAN and etc)
+    pub fn set_filter_mask(mut self, family: u8, filt_mask: u32) -> Self {
```
little-dude, line 38: This is kind a bit redundant with `new()`, could we make this simply:

```rust
pub fn set_filter_mask(mut self, family: u8, filt_mask: u32) -> Self {
     self.message.header.interface_family = family;
     self.message.nlas.push(Nla::ExtMask(filter_mask));
     self
}
```

Then it would be called like:

```rust
LinkGetRequest::new(handle)
    .set_filter_mask(family, filt_mask)
    .execute()
```

Also, do we necessarily need the address family when setting a filter?
Or should this be a different `set_interface_family()` method?

rtnetlink/Cargo.toml
```diff
@@ -17,6 +17,7 @@ log = "0.4.8"
 thiserror = "1"
 netlink-packet-route = { path = "../netlink-packet-route", version = "0.2" }
 netlink-proto = { path = "../netlink-proto", version = "0.2" }
+netlink-sys = {path = "../netlink-sys", version = "0.2" }
```
JohnTitor, line 20: This should be dev-dep.

rtnetlink/src/link/get.rs
cathay4t: Changing the `LinkGetRequest::new()` would break the API which I think we can discuss that in another pull request while this PR focusing on extending existing API for `filter mask`.
But if you prefer to do it in a bundle, I am OK with it, just let me know.

cathay4t: I tested several days ago, it is required for vlan filtering, but let me test again.

cathay4t: It it required to set `interface_family`. Without it, the `AfBridgeSpec` will not show. Tested on kernel 5.4.43(archlinux).

For kernel code, the `validate_linkmsg()` in `net/core/rtnetlink.c` will ensure the returned AfSpec interface family matches, and `br_fill_ifinfo()` in `net/bridge/br_netlink.c` set the `hdr->ifi_family = AF_BRIDGE`.

cathay4t:
New patch uploaded.

Changes since last one:
 * Changed to in place update without `new()`.
 * Fixed the `rtnetlink/Cargo.toml` to include the new dependency as dev only.

rtnetlink/src/link/get.rs
little-dude: Sounds good, thank you.

little-dude:
@cathay4t do you want me to release new crates with all your recent changes, or do you have some more you might push?

cathay4t:
@little-dude I don't have work in progress[0]. Release a new crates could helps. So that my [project][1] could depend on.

[1]: https://github.com/nispor/nispor

little-dude:
Cool. Will do tonight!

cathay4t:
@little-dude Any plan for the new release?

little-dude:
Sorry I forgot. I just released these new versions:

- audit 0.2.1
- rtnetlink 0.3.0
- netlink-packet-audit 0.2.0
- netlink-proto 0.4.0 (should have been 0.3.0 but I messed up)
- netlink-packet-route 0.3.0
- netlink-packet-core 0.2.0
- netlink-packet-utils 0.2.0
- netlink-sys 0.3.0

PR #81, opened 2020-05-29
Fix endian of IFLA_VLAN_PROTOCOL

cathay4t:
The IFLA_VLAN_PROTOCOL is stored as big endian u16, the kernel code is:
`net/8021q/vlan_netlink.c`:

    if (nla_put_be16(skb, IFLA_VLAN_PROTOCOL, vlan->vlan_proto) ||

Test case fixed also.


little-dude:
Thank you for all these fixes @cathay4t !

PR #82, opened 2020-05-29
Tweak travis config

JohnTitor:
- clippy -> seems they're duplicate so shrunk to one
- rustfmt -> another option is to use stable one since nightly has unstable formatting
- allow nightlies to fail correctly
- fix/suppress warnings


little-dude:
Very nice, this was long overdue indeed, so thank you @JohnTitor

PR #85, opened 2020-07-01
rtnetlink: Add support of querying qdisc in traffic control

cathay4t:
New function `rtnetlink::Handle::qdisc()` with initial qdisc query
support.

Integration test case added which query the qdisc information of
loopback interface.


cathay4t:
@little-dude I failed to find example on testing this new handler, do you how should I create the test case for this new functionality?

little-dude:
It would be nice to have an example yes. You can re-use an existing one or create a new one, I don't really have a preference.

cathay4t:
@little-dude Integration test case added which test on qdisc of loopback interface and patch is ready for review.

cathay4t:
Still work in progress and my project need more properties been exposed.

PR #86, opened 2020-07-08
Audit: add get_status method

idanski:
It seems to work and return the actual audit status.

I'll be happy for a review :grin: 

Thanks!!

### DIFF
--- a/audit/src/handle.rs
+++ b/audit/src/handle.rs
@@ -121,4 +121,27 @@ impl Handle {
         req.header.flags = NLM_F_REQUEST | NLM_F_ACK;
         self.acked_request(req).await
     }
+
+    /// Get current audit status
+    pub async fn get_status(&mut self) -> Result<StatusMessage, Error> {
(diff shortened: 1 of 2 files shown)


audit/src/handle.rs
```diff
@@ -121,4 +121,27 @@ impl Handle {
...
         self.acked_request(req).await
     }
+
+    /// Get current audit status
+    pub async fn get_status(&mut self) -> Result<StatusMessage, Error> {
+        let mut req = NetlinkMessage::from(AuditMessage::GetStatus(None));
+        req.header.flags = NLM_F_REQUEST | NLM_F_DUMP;
+
+        let mut request = self.request(req)?;
+        if let Some(response) = request.next().await {
```
r-darwish, line 131: I think you can do `let response = request.next().await.ok_or(Error::RequestFailed)?` and get rid of the `if let`

little-dude:
Thank you!

PR #88, opened 2020-07-26
rtnetlink: Add support of querying classes in traffic control

cathay4t:
New function `rtnetlink::Handle::traffic_class()` with initial traffic
class query support.

Integration test case added which query the traffic class information of
loopback interface.
By default, there is no traffic class on any NICs, to test it, we need
to add one and do the query. So `cargo test` need to be run as root or
user with enough privilege for `tc class add` command.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,10 @@ matrix:
       script:
         - cd rtnetlink
         - cargo test
+        - sudo env PATH=${TRAVIS_HOME}/.cargo/bin:$PATH
+          cargo test --features test_as_root
+        - sudo chown -R travis:travis $HOME/.cargo
+        - sudo chown -R travis:travis $TRAVIS_BUILD_DIR/target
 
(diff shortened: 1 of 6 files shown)


little-dude:
Hi @cathay4t sorry for being a bit slow on this, I was on vacations. I only skimmed through the code but that looks great to me codewise. However I'd like to avoid requiring root privileges for running the tests. Maybe it's possible to keep such tests behind a `test-as-root` feature flag for example? Although I don't know if we can have feature flags for just for the tests...

cathay4t:
Will take a look.

cathay4t:
New patch uploaded.

The newly add test case is ignored by default in command `cargo test`.
It can be enabled by `sudo cargo test --features test_as_root` or
`sudo cargo test --all-features`.

little-dude:
Thanks a lot @cathay4t :)

PR #89, opened 2020-07-30
Add Socket. recv_from_full

Tim-Zhang:
This method will first peek to obtain the length, and then receive a message of this length.

### DIFF
--- a/netlink-sys/src/sys.rs
+++ b/netlink-sys/src/sys.rs
@@ -263,6 +263,21 @@ impl Socket {
         Ok(res as usize)
     }
 
+    /// Receive a full message.
+    /// Unlike recv_from, which truncates messages that exceed the length of the buffer passed as argument,
+    /// this method always reads a whole message, no matter its size.
+    pub fn recv_from_full(&self) -> Result<(Vec<u8>, SocketAddr)> {
+        // Peek
+        let mut buf = Vec::<u8>::new();
+        let (rlen, _) = self.recv_from(&mut buf, libc::MSG_PEEK | libc::MSG_TRUNC)?;
+
+        // Receive
+        let mut buf = vec![0; rlen as usize];
+        let (_, addr) = self.recv_from(&mut buf, 0)?;
+
+        Ok((buf, addr))
+    }
+
     pub fn send_to(&self, buf: &[u8], addr: &SocketAddr, flags: libc::c_int) -> Result<usize> {
         let (addr_ptr, addr_len) = addr.as_raw();
         let buf_ptr = buf.as_ptr() as *const libc::c_void;
--- a/netlink-sys/src/tokio.rs
+++ b/netlink-sys/src/tokio.rs
@@ -102,6 +102,10 @@ impl Socket {
         poll_fn(|cx| self.poll_recv_from(cx, buf)).await
     }
 
+    pub async fn recv_from_full(&mut self) -> io::Result<(Vec<u8>, SocketAddr)> {
+        poll_fn(|cx| self.poll_recv_from_full(cx)).await
+    }
+
     pub fn poll_recv_from(
         &mut self,
         cx: &mut Context,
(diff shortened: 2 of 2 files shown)


netlink-sys/src/sys.rs
```diff
@@ -263,6 +263,51 @@ impl Socket {
...
+            rlen = libc::recvmsg(self.0, &mut h as *mut libc::msghdr, 0);
+            if rlen < 0 {
+                return Err(Error::last_os_error());
+            }
+
+            if h.msg_flags & libc::MSG_TRUNC != 0 {
+                return Err(std::io::ErrorKind::InvalidData.into());
+            }
+
+            v.resize(rlen as usize, 0);
```
little-dude, line 305: Is it really possible for `rlen` to be different from when we peeked the message?

Tim-Zhang:
Hi, @little-dude 

I have updated the code to remove `recvmsg`, please review it again. 

Thank you

netlink-sys/src/sys.rs
```diff
@@ -263,6 +263,18 @@ impl Socket {
         Ok(res as usize)
     }
 
+    pub fn recv_all(&self) -> Result<(Vec<u8>, SocketAddr)> {
```
little-dude, line 266: If I understand correctly, unlike `recv_from` which can truncate the message, `recv_all` will return the full message, correct? It would be nice to document that behavior by adding a docstring to both `recv_from` and `recv_all`.


Also what do you think about renaming `recv_all` into something like `recv_from_full` or `recv_from_no_truncate`? It doesn't looks as nice as `recv_all` but it makes it clearer that this is basically a variant of `recv_from`, imo.

Tim-Zhang: @little-dude  I think these two names are not good names, even `recv_from` is. 
`recv_from` from where/who?  the address pointer is not exposed, right? 
I think these names should not rely on underlying libc function name, we just name it by its action. maybe one day we could replace all libc::recvfrom to libc::recvmsg without rename the function

How about rename `recv_all` to `recv_all_available` avoid confusion with `recv_to_end`?

Tim-Zhang:
Hi @little-dude Is there any progress?

netlink-sys/src/sys.rs
little-dude: `recv_from` comes from the `recvfrom` C API. Sure we don't pass a function pointer as an argument, but we do return the sender's address.

```
ssize_t recv(int sockfd, void *buf, size_t len, int flags);
```

becomes 

```
fn recv(&self, buf: &mut [u8], flags: libc::c_int) -> Result<usize>
``` 

and 

```
ssize_t recvfrom(int sockfd, void *buf, size_t len, int flags,
                 struct sockaddr *src_addr, socklen_t *addrlen);
```

becomes

```
pub fn recv_from(&self, buf: &mut [u8], flags: libc::c_int) -> Result<(usize, SocketAddr)>
```

The function that you introduced is basically a form of `recvfrom` so that's why I'd like to keep that naming.

little-dude:
Hi @Tim-Zhang I apologize for taking so long with this...

netlink-sys/src/sys.rs
Tim-Zhang: well, I will rename it to recv_from_full just hope to merge quickly.

Tim-Zhang:
Hi @little-dude, I have changed the name, please review it again, thank you.

netlink-sys/src/sys.rs
```diff
@@ -263,6 +263,19 @@ impl Socket {
         Ok(res as usize)
     }
 
+    /// Receive all available messages
```
little-dude, line 266: Shouldn't this be something like:

> Receive a full message. Unlike `recv_from`, which truncates messages that exceed the length of the buffer passed as argument, this method always reads a whole message, no matter its size.

little-dude (approved):
Thanks. Please let me know what you think of the suggested docstring. As soon as we agree on that I'll merge.

little-dude:
I guess you'll need a release once this is merged?

netlink-sys/src/sys.rs
Tim-Zhang: done

Tim-Zhang:
> I guess you'll need a release once this is merged?

Yes, I need a release, should I update the version in this pr ?

little-dude:
Release netlink-sys 0.4 @Tim-Zhang  :)

Tim-Zhang:
@little-dude Thank you very much!

PR #91, opened 2020-08-02
rtnetlink: Add support of querying filters in traffic control

cathay4t:
New function `rtnetlink::Handle::traffic_filter()` with initial traffic
filter query support.

The `cargo test` will run in parallel, changing loopback interface might
cause `test_get_qdiscs` fail.
Hence changed the integration test case to use dummy interface instead
loopback.

### DIFF
--- a/rtnetlink/src/traffic_control/get.rs
+++ b/rtnetlink/src/traffic_control/get.rs
@@ -58,10 +58,7 @@ impl TrafficClassGetRequest {
     pub(crate) fn new(handle: Handle, ifindex: i32) -> Self {
         let mut message = TcMessage::default();
         message.header.index = ifindex;
-        TrafficClassGetRequest {
-            handle,
-            message,
-        }
+        TrafficClassGetRequest { handle, message }
     }
 
     /// Execute the request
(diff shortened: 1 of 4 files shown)


little-dude:
> The cargo test will run in parallel, changing loopback interface might
cause test_get_qdiscs fail.

We can also run the tests with `cargo test -- --test-threads=1` if that makes things easier.

rtnetlink/src/traffic_control/get.rs
```diff
@@ -90,3 +87,42 @@ impl TrafficClassGetRequest {
         }
     }
 }
+
+pub struct TrafficFilterGetRequest {
```
little-dude, line 91: I'm not too familiar with `tc`, so would you mind adding a docstring with the equivalent `tc` commands this request corresponds to?

cathay4t:
> > The cargo test will run in parallel, changing loopback interface might
> > cause test_get_qdiscs fail.
> 
> We can also run the tests with `cargo test -- --test-threads=1` if that makes things easier.

Then we have to explain in document said `cargo test` will fail. Not too complex to work around it.
So let's just keep it in this way.

rtnetlink/src/traffic_control/get.rs
cathay4t: The document is at `Handle::traffic_filter()`:

```
    /// Create a new handle, specifically for traffic control filter requests
    /// (equivalent to `tc filter show dev <interface_name>` commands)
```

PR #95, opened 2020-08-21
Impl AsRawFd for tokio Socket

Tim-Zhang:
Provide a method to get the fd so that the fd can be manipulated.
e.g. fcntl(fd, xxx)


little-dude:
Thank you!

PR #96, opened 2020-08-26
Include license file in the crate

cathay4t:
The crate file created by `cargo package` does not contains
license file.

Create soft-link to each crate and `cargo package` will include the
real license file in the crate file.


little-dude:
Thanks @cathay4t. Should I bump the version of each of these crates?

cathay4t:
No need for this trivial.

Issue #27, opened 2019-01-02
implement `ip monitor`


little-dude:
This should not be too hard. All we need to bind the netlink socket to multicast groups we're interested in. We already do something similar for audit events (see https://github.com/little-dude/netlink/blob/master/audit/examples/events.rs), and most of the code for implementing this already exists in netlink-proto.

PR #97, opened 2020-09-05
ip monitor clone

SkamDart:
Took a stab at #27 
Very good chance I am missing something but the gist of it is from
[iproute2](https://github.com/shemminger/iproute2/blob/f33a871b8094ae0f6e6293804e1cc6edbba0e108/ip/ipmonitor.c#L168-L355)


stappersg (approved):
Works for me

PR #56, opened 2019-11-03
WIP: sock_diag

little-dude:
Taking over https://github.com/little-dude/netlink/pull/32.
I think this is going to take a couple weeks.

PR #99, opened 2020-09-23
sock_diag: wire up

george-hopkins:
As I needed to quickly diagnose some sockets, I took the liberty to finish up your work in #56. Specifically, I added a buffer for the messages and an example. Other than a small fix of the serialization of port numbers, it's just a rebased version of your implementation. I'm happy to close this PR and rebase my fixes on your work in case you'd like to keep working on it.

### DIFF
--- a/netlink-packet-sock-diag/Cargo.toml
+++ b/netlink-packet-sock-diag/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+authors = ["Flier Lu <flier.lu@gmail.com>", "Corentin Henry <corentinhenry@gmail.com>"]
+name = "netlink-packet-sock-diag"
+version = "0.1.0"
+edition = "2018"
+
+homepage = "https://github.com/little-dude/netlink"
+keywords = ["netlink", "linux", "sock_diag"]
+license = "MIT"
+readme = "../README.md"
+repository = "https://github.com/little-dude/netlink"
+description = "netlink packet types for the sock_diag subprotocol"
+
+[dependencies]
+failure = "0.1.5"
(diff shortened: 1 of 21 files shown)


little-dude:
Hi @george-hopkins this is very cool thank you! I'll take a look.

> I'm happy to close this PR and rebase my fixes on your work in case you'd like to keep working on it.

Well at this point there have already been two attempts to land this. The problem is that it's massive. So maybe we should change strategy and merge a first implementation even if it's incomplete, and work from there. So I'm tempted to try to merge this already. Wdyt?

george-hopkins:
Thank you, it's quite a big addition indeed. However, I think it can serve as a good base due to the good architecture which handles unparsed attributes quite nicely. In addition, I think it will lower the bar for improvements and encourage other developers to contribute the remaining bits over time if the base is already laid out and can serve as an example.

little-dude:
All right. Let merge this in then :)

netlink-packet-sock-diag/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+authors = ["Flier Lu <flier.lu@gmail.com>", "Corentin Henry <corentinhenry@gmail.com>"]
```
little-dude, line 2: @george-hopkins I'd like to release 0.1, and noticed that you haven't added yourself to the authors. Would you like to? If so which name/email would you like to use?

george-hopkins: Thank you for asking! It's fine by me, I'm glad `sock_diag` is officially supported now :tada:

PR #100, opened 2020-09-28
netlink-packet-route: Implement IPvlan LinkInfo NLA

nightspc:
Tried my hand at implementing IPvlan LinkInfo handling.
I had wanted to add the ipvlan_mode enum and IPvlan flag constants defined in linux/include/uapi/linux/if_link.h but I wasn't sure where to put them.

As an aside, the tests I added ended up uncovering a bug in my iproute2 as it was sending link info where the info kind field length was 1 byte shorter than it should be compared to both this library and the kernel's response. Not sure if this is fixed upstream (found in v 5.8.0).

### DIFF
--- a/netlink-packet-route/src/rtnl/constants.rs
+++ b/netlink-packet-route/src/rtnl/constants.rs
@@ -443,6 +443,9 @@ pub const IFLA_VLAN_FLAGS: u16 = 2;
 pub const IFLA_VLAN_EGRESS_QOS: u16 = 3;
 pub const IFLA_VLAN_INGRESS_QOS: u16 = 4;
 pub const IFLA_VLAN_PROTOCOL: u16 = 5;
+pub const IFLA_IPVLAN_UNSPEC: u16 = 0;
+pub const IFLA_IPVLAN_MODE: u16 = 1;
+pub const IFLA_IPVLAN_FLAGS: u16 = 2;
 pub const VETH_INFO_UNSPEC: u16 = 0;
 pub const VETH_INFO_PEER: u16 = 1;
 
--- a/netlink-packet-route/src/rtnl/link/nlas/link_infos.rs
+++ b/netlink-packet-route/src/rtnl/link/nlas/link_infos.rs
@@ -1330,6 +1330,58 @@ mod tests {
         assert_eq!(expected, parsed);
     }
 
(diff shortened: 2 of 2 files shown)


little-dude:
Thanks a lot for the PR @nightspc !

> As an aside, the tests I added ended up uncovering a bug in my iproute2 as it was sending link info where the info kind field length was 1 byte shorter than it should be compared to both this library and the kernel's response. Not sure if this is fixed upstream (found in v 5.8.0).

Two things I can think of:

- `iproute2` is sometimes inconsistent with null terminated strings: sometimes strings are null terminated, sometimes they are not. I tried to be flexible here, so [`parse_string`](https://github.com/little-dude/netlink/blob/5e328e340a2a729269552968217b6b39dc082a08/netlink-packet-utils/src/parsers.rs#L35-L39) tries to be smart. [Here is a test for this](https://github.com/little-dude/netlink/blob/252822f8c72e497bfcbe7794eac963b089d9b7ad/netlink-packet-route/src/rtnl/test.rs#L16)
- I've noticed that sometimes, the padding is not counted. I thought I had a test for this as well but I cannot find it...

If you could share the message, I'd love to add a test for it, because I'd like this crate to play nice with `iproute2`

little-dude:
I released 0.5 with your changes.

nightspc:
Thank you for merging @little-dude !

As for the link creation message:
```
0000   54 00 00 00 10 00 05 06 52 85 73 5f 00 00 00 00   T.......R.s_....
0010   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0020   08 00 05 00 02 00 00 00 08 00 03 00 69 76 30 00   ............iv0.
0030   24 00 12 00 0a 00 01 00 69 70 76 6c 61 6e 00 00   $.......ipvlan..
0040   14 00 02 00 06 00 01 00 01 00 00 00 06 00 02 00   ................
0050   02 00 00 00                                       ....
```
byte 5 on line 0030 (0a) is the length of the field.
Full netlink capture [see packet 5]: https://pubfiles.replicantprocess.com/create-iv0-ipvlan.pcap

Messages were created by running `ip link add iv0 type ipvlan link dummy0 mode l3 vepa`

Hope that helps you!

PR #110, opened 2020-10-04
Set up for GitHub Actions

### LINKED ISSUE #106: Move to Github Actions
Travis jobs take ages to start, and I find Github Actions easier to maintain.

JohnTitor:
Fixes #106 

This migrates our CI to GHA with some code fixes, seems travis couldn't find them. You could find the log in https://github.com/JohnTitor/netlink/actions.


netlink-sys/src/addr.rs
```diff
@@ -64,7 +64,7 @@ use std::{
 /// The application can also pick the port number by calling Socket::bind with an address with a non-zero port
 /// number. However, it must ensure that this number is unique for each socket created. For instance:
 ///
-/// ```rust
+/// ```rust,should_panic
```
little-dude, line 67: mmh this is not supposed to panic, do you mind sharing the error you get?

little-dude:
Awesome, thanks a lot @JohnTitor. Seems like the actions are not triggered though, is it because it needs to be merged first?

.github/workflows/main.yml
```diff
@@ -0,0 +1,73 @@
...
+          rustc -vV
+          rustup update stable
+          rustup default stable
+          rustc -vV
+
+      - name: test (netlink-sys)
+        run: |
+          cd netlink-sys
+          cargo test
+          cargo test --features tokio_socket
```
little-dude, line 32: Not necessarily to be addressed in this PR but `cargo test --features smol_socket` currently fails. We should fix this and add it to CI afterward.

little-dude:
@JohnTitor I'm leaving for the airport and we'll probably be AFK for the whole week, so I'll let you merge this :) Thanks again !

netlink-sys/src/addr.rs
JohnTitor: Hm I think this causes an error about "address already in use", doesn't it?

https://github.com/little-dude/netlink/blob/d799194a13c653c440652e1fe44e03a7f72b03bb/netlink-sys/src/addr.rs#L79-L82

JohnTitor:
> Seems like the actions are not triggered though, is it because it needs to be merged first?

Yeah, PR coming from fork repo doesn't trigger GHA for a security reason AFAIK.

> I'm leaving for the airport and will probably be AFK for the whole week, so I'll let you merge this :)

I'd wait as the doc question isn't resolved and I think this isn't urgent :)

netlink-sys/src/addr.rs
little-dude: Yes, the "address already in use" is expected, but that's why we `unwrap_err` actually.

```rust
socket2.bind(&addr2).unwrap_err(); 
```

My only hypothesis is that the first `bind` could trigger this error, if we have another test running in parallel that uses the same address. Could you try with `cargo test -- --test-threads=1`?

JohnTitor: Oh, I totally missed `unwrap_err`, sorry!
So after some debugging, the `test-threads` flag didn't work well (https://github.com/JohnTitor/netlink/runs/1240889137?check_suite_focus=true), and using socket number "2" worked as expected (https://github.com/JohnTitor/netlink/runs/1240969349?check_suite_focus=true). Uhm, generally, I'd avoid changing docs just only for CI but it's quite minor. If you don't want to change it, I'd add the `--lib` flag to exclude doctests until we find a better way.

little-dude: Thanks for all the debugging. I think changing the docs here is fine. I'm a little worried about not understanding where the failure comes from but I'll do some more debugging later.

little-dude (approved):
Thanks a lot @JohnTitor. I'm not too familiar with GA myself, so this is awesome :+1:

netlink-sys/src/addr.rs
JohnTitor: Okay, removed all the debug commits and this is now ready to merge :)

> I'm a little worried about not understanding where the failure comes from but I'll do some more debugging later. It passes on my local machines (nixos and archlinux) and on Travis though, so maybe Github Actions just do something weird.

The failure comes from the first `bind`'s unwrap so I guess the port number is used on GHA anyways :thinking:

PR #102, opened 2020-10-01
netlink-packet-route: support IP over Infiniband interface

ffmancera:
This patch is introducing IP over Infiniband interface support. In addition, it is providing the InfoIpoib Nla. This Nla is gathering all the IFLA_IPOIB_ netlink attributes.

Signed-off-by: Fernando Fernandez Mancera <ffmancera@riseup.net>

PR #111, opened 2020-10-04
netlink-packet-route: Implement MacVlan LinkInfo NLA

nightspc:
This adds support for MacVlan Link infos.

Notably, this adds support for the source mode of MacVlan interfaces, which allows specifying a set of Mac Addresses. This is implemented by packing MacVlan MacAddr NLA's into the MacAddr Data NLA.
My implementation of this calls the MacVlan parser recursively to generate these but this allows other MacAddr NLA's to be used in this field which may not be wanted behavior as the kernel will reject anything except MacVlan MacAddr NLAs.

Additionally, Unspec parsing is implemented but the conversation about whether this is appropriate in #102 was never resolved so I've included it here.

Finally, the same null byte issue from #100 is present in iproute2 when using macvlan, causing this library to have different behavior to it.


little-dude:
Hi @nightspc!

Thanks for the PR, and sorry for being a bit late on the review, I was on vacation last week.

> My implementation of this calls the MacVlan parser recursively to generate these but this allows other MacAddr NLA's to be used in this field which may not be wanted behavior as the kernel will reject anything except MacVlan MacAddr NLAs.

I think this is fine. There are other places where we already do something similar.

little-dude (approved):
Thanks for adding tests for this :)

flouthoc:
@nightspc @little-dude Hi thanks a lot for this great addon and awesome work.

It seems `netlink` is still missing support for `macvlan` with `LinkAddRequest`.  Could i make this work with minimal changes ?

Issue #113, opened 2020-11-16
Adding VXLAN Support


gabrik:
Hi, 

First, thank you for this very useful crate.

I noticed that in `rtnetlink` only some types of interfaces can be created.
And I also see that the VxLAN info is not yet defined: https://github.com/little-dude/netlink/blob/e9ece979dc134fc0080c84f1fcf34a730a3a52bd/netlink-packet-route/src/rtnl/link/nlas/link_infos.rs#L215 

As this is very interesting for me to have if you can point me to some references I can try to add it and do a PR with the code.

Should I follow libnl for the definition of the attributes? https://www.infradead.org/~tgr/libnl/doc/api/vxlan_8c_source.html 

Thanks

little-dude:
Hey, thanks a lot for taking a stab at this!
I think `libnl` is indeed a good source. The netlink go library seems to also support vxlan attributes, so you may look there as well, if you want: https://github.com/vishvananda/netlink/blob/d40f9887b852ec4ce633f2a46734422c87fc8336/link_linux.go#L904

gabrik:
This is was implemented by #114, I'm closing this issue

PR #114, opened 2020-11-17
Implementing Vxlan 

gabrik:
Hi,

This PR will contain the code to implement the VXLAN interface creation in `rtnetlink`
Will solve https://github.com/little-dude/netlink/issues/113 
- [x] create InfoVxlan enum
- [x] impl Nla for InfoVxlan
- [x] impl Parseable for InfoVxlan
- [x] add missing fields in InfoVxlan enum
- [x] add vxlan parsing in VecInfo::parse
- [x] add vxlan function to `rtnetlink` add

It is made as a draft so I can keep track of the changes


gabrik:
@little-dude 

I have almost everything in place, I also added an example (not committed yet) but I see netlink returning an error.


```
./target/debug/examples/create_vxlan eno0
Error: "Received a netlink error message Numerical result out of range (os error 34)
```

I guess I need to pass some other netlink attributes, I guess I can check with libnl the ones that are required right?

little-dude:
Hi @gabrik 

You error seems to suggest that one attribute is set to the wrong value, not that you're missing an attribute: `Numerical result out of range` (I should fix the error message punctuation so that is looks like `Received a netlink error message: Numerical result out of range (os error 34)`).

The way I proceed for this kind of things is usually to find out how the `ip` command does it (which attributes it sets etc.). To see the messages sent by the `ip` command you basically have the choice:

- capture netlink packets and analyze them in wireshark
- or use `strace` to see the netlink packagets in the terminal

These two techniques are explained in more details in this blog post: https://jvns.ca/blog/2017/09/03/debugging-netlink-requests/

gabrik:
Solved, at the end was an endianness issue, thanks to `strace` I discovered that while I was sending big-endian, `ip link` was sending little-endian.
Now I'm using NativeEndian.

I guess it is ready for review.

The vxlan creation function is very stupid just takes the vni and the link, so I'm not 100% happy, any suggestion to improve it is more than welcome.

Ideally I would like something like this:

```
...
.link()
.add()
.vxlan("vxlan0".into(), link.header.index, 1u32)
.group(...)
.ttl(...)
.port(...)
.execute()
```

What do you think?

little-dude:
Thanks a lot @gabrik ! I'll get this reviewed tonight (european time).

netlink-packet-route/src/rtnl/link/nlas/link_infos.rs
```diff
@@ -437,6 +456,229 @@ impl<'a, T: AsRef<[u8]> + ?Sized> Parseable<NlaBuffer<&'a T>> for InfoKind {
...
+    //TtlInherit
+    Df(u8),
+}
+
+impl Nla for InfoVxlan {
+    #[rustfmt::skip]
+    fn value_len(&self) -> usize {
+        use self::InfoVxlan::*;
+        match self {
+            Tos(_) |
```
little-dude, line 499: I usually put the `|` before:
```
match self {
    Tos(_)
        | Tls(_)
        |
// etc.
```

rtnetlink/examples/create_vxlan.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    create_vxlan(handle, link_name.to_string())
+        .await
+        .map_err(|e| format!("{}", e))
+}
+
+async fn create_vxlan(handle: Handle, name: String) -> Result<(), Error> {
+    let mut links = handle.link().get().set_name_filter(name.clone()).execute();
+    if let Some(link) = links.try_next().await? {
+        handle
```
little-dude, line 25: That looks pretty nice imo :+1:

rtnetlink/src/link/add.rs
```diff
@@ -17,6 +17,137 @@ use crate::{
...
+
+    pub fn group6(mut self, addr: std::net::Ipv6Addr) -> Self {
+        let ip: u128 = u128::from_le_bytes(addr.octets());
+        self.info_data.push(InfoVxlan::Group6(ip));
+        self
+    }
+
+    pub fn local(mut self, addr: std::net::Ipv4Addr) -> Self {
+        let octets = addr.octets();
+        let ip: u32 = (octets[0] as u32) << 24
```
little-dude, line 70: Maybe:
```rust
u32::from_bytes_be(addr.octets())
```
(if I got my endianness right)

rtnetlink/src/link/add.rs
```diff
@@ -17,6 +17,137 @@ use crate::{
...
+        self.info_data.push(InfoVxlan::Link(index));
+        self
+    }
+
+    pub fn port(mut self, port: u16) -> Self {
+        self.info_data.push(InfoVxlan::Port(port));
+        self
+    }
+
+    pub fn group(mut self, addr: std::net::Ipv4Addr) -> Self {
```
little-dude, line 52: Would you mind addind some doc to these methods? It doesn't need to be an in-depth explanation of VXLAN, just a couple sentences would be nice, or the equivalent `ip` command.

rtnetlink/src/link/add.rs
```diff
@@ -117,6 +248,17 @@ impl LinkAddRequest {
             .up()
     }
 
+    /// Create a VxLAN on a link
+    /// This is equivalent to ip link add name NAME type vxlan id VNI dev LINK
```
little-dude, line 252:
```suggestion
    /// This is equivalent to `ip link add name NAME type vxlan id VNI dev LINK`
```

little-dude (commented):
The changes look good, thanks a lot @gabrik.

I'd just like to have some documentation in the `rtnetlink` crate if you don't mind.

rtnetlink/src/link/add.rs
gabrik: I was thinking that maybe instead of using having hard-coded BE or LE, I would like to check the machine endianness and use the correct one.
I was looking at this: https://docs.rs/byteordered/0.5.0/byteordered/enum.Endianness.html 

If you are ok with this I can use that crate as a dependency and use always the correct endianness

little-dude:
Thanks a lot @gabrik !

Issue #38, opened 2019-02-08
Implement network namespace creation a la iproute2


little-dude:
Some workflows with the route protocol involve configuring interfaces in other network namespaces. However, the namespace creation itself is not part of the netlink protocol. One must use the `unshare()` or `clone()` syscalls for that. 

Another problem is keeping such namespaces alive even if no process uses them: if no entry in `/proc` points to a given namespace, the kernel deletes it. `iproute2` solves this by creating another mount point under `/var/run/` and mounting the namespace filesystem there to keep it alive (see https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/ip/ipnetns.c#n661).

We could try to do the same, but I'm not sure to fully understand all the details of the code yet especially this specific loop:

```c
	/* Make it possible for network namespace mounts to propagate between
	 * mount namespaces.  This makes it likely that a unmounting a network
	 * namespace file in one namespace will unmount the network namespace
	 * file in all namespaces allowing the network namespace to be freed
	 * sooner.
	 */
	while (mount("", NETNS_RUN_DIR, "none", MS_SHARED | MS_REC, NULL)) {
		/* Fail unless we need to make the mount point */
		if (errno != EINVAL || made_netns_run_dir_mount) {
			fprintf(stderr, "mount --make-shared %s failed: %s\n",
				NETNS_RUN_DIR, strerror(errno));
			return -1;
		}

		/* Upgrade NETNS_RUN_DIR to a mount point */
		if (mount(NETNS_RUN_DIR, NETNS_RUN_DIR, "none", MS_BIND | MS_REC, NULL)) {
			fprintf(stderr, "mount --bind %s %s failed: %s\n",
				NETNS_RUN_DIR, NETNS_RUN_DIR, strerror(errno));
			return -1;
		}
		made_netns_run_dir_mount = 1;
	}
```

A second issue would be to integrate this with Tokio: `mount` is blocking (not sure if it blocks the whole _process_ or just the _thread_), so it will block the entire tokio event loop, so we should maybe use a separate thread for that.

All the code for this should probably be put in a separate crate.

little-dude:
Here is an `strace` capture of `ip netns add blue` for the record`:

```rust
==== open the netlink socket                                                                                                                                                                                                                                                                                    
socket(AF_NETLINK, SOCK_RAW|SOCK_CLOEXEC, NETLINK_ROUTE) = 3                                                                                                                                                                                                                                                    
[...]

PR #116, opened 2020-11-20
Implementing network namespace addition and removal a la ip netns

gabrik:
I will use this PR to keep track of the implementation of the creation and deletion of network namespaces in a similar way as done by `ip netns` command (#38).

- [x] Async network namespace creation
- [x] Async network namespace deletion
- [x] example
- [x] docs

### DIFF
--- a/rtnetlink/Cargo.toml
+++ b/rtnetlink/Cargo.toml
@@ -23,6 +23,7 @@ netlink-packet-route = { path = "../netlink-packet-route", version = "0.6" }
 netlink-proto = { path = "../netlink-proto", version = "0.5" }
 byteordered = "0.5.0"
 nix = "0.19.0"
+tokio = { version = "0.2.6", features = ["full"] }
 
 [dev-dependencies]
 env_logger = "0.7.1"
--- a/rtnetlink/Cargo.toml
+++ b/rtnetlink/Cargo.toml
@@ -22,6 +22,7 @@ thiserror = "1"
 netlink-packet-route = { path = "../netlink-packet-route", version = "0.6" }
 netlink-proto = { path = "../netlink-proto", version = "0.5" }
 byteordered = "0.5.0"
+nix = "0.19.0"
 
 [dev-dependencies]
 env_logger = "0.7.1"
--- a/rtnetlink/src/ns.rs
+++ b/rtnetlink/src/ns.rs
@@ -1,0 +1,77 @@
+use nix::fcntl::OFlag;
+use nix::sched::*;
+use nix::sys::stat::Mode;
+use std::option::Option;
+use std::path::Path;
+
+pub const NETNS_PATH: &str = "/var/run/netns/";
+pub const SELF_NS_PATH: &str = "/proc/self/ns/net";
+pub const NONE_FS: &str = "none";
+
+
+pub struct NetworkNamespace();
+
+
+impl NetworkNamespace {
+    pub async fn add(ns_name: &String) -> nix::Result<()> {
+        let mut netns_path = String::new();
+
+        let dir_path = Path::new(NETNS_PATH);
+        let mut mkdir_mode = Mode::empty();
+        let mut open_flags = OFlag::empty();
+
+        // flags in mkdir
+        mkdir_mode.insert(Mode::S_IRWXU);
+        mkdir_mode.insert(Mode::S_IRGRP);
+        mkdir_mode.insert(Mode::S_IXGRP);
+        mkdir_mode.insert(Mode::S_IROTH);
+        mkdir_mode.insert(Mode::S_IXOTH);
+
+        open_flags.insert(OFlag::O_RDONLY);
+        open_flags.insert(OFlag::O_CREAT);
+        open_flags.insert(OFlag::O_EXCL);
+
+        netns_path.push_str(NETNS_PATH);
+        netns_path.push_str(ns_name);
+
+        // creating namespaces folder if not exists
+        nix::unistd::mkdir(dir_path, mkdir_mode)?;
+
+        let ns_path = Path::new(&netns_path);
+
+        // creating the netns file
+        let fd = nix::fcntl::open(ns_path, open_flags, Mode::empty())?;
+
+        nix::unistd::close(fd)?;
+
+        // unshare to the new network namespace
+        let ret = nix::sched::unshare(CloneFlags::CLONE_NEWNET);
+
+        let self_path = Path::new(&SELF_NS_PATH);
+        let none_fs = Path::new(&NONE_FS);
+        let none_p4: Option<&Path> = None;
+        // bind to the netns
+        nix::mount::mount(
+            Some(self_path),
+            ns_path,
+            Some(none_fs),
+            nix::mount::MsFlags::MS_BIND,
+            none_p4,
+        )?;
+
+        ret
+    }
+
+
+    pub async fn del(ns_name: &String) -> nix::Result<()> {
+        let mut netns_path = String::new();
+        netns_path.push_str(NETNS_PATH);
+        netns_path.push_str(ns_name);
+        let ns_path = Path::new(&netns_path);
+
+        nix::mount::umount2(ns_path, nix::mount::MntFlags::MNT_DETACH)?;
+
+        nix::unistd::unlink(ns_path)
+    }
+
+}
--- a/rtnetlink/src/ns.rs
+++ b/rtnetlink/src/ns.rs
@@ -194,7 +194,7 @@ impl NetworkNamespace {
     /// Remove a network namespace
     /// This is equivalent to `ip netns del NS_NAME`.
     pub async fn del(ns_name: String) -> Result<(), Error> {
-        let jh = task::spawn_blocking(move || {
+        let res = task::spawn_blocking(move || {
             let mut netns_path = String::new();
             netns_path.push_str(NETNS_PATH);
             netns_path.push_str(&ns_name);
@@ -212,7 +212,7 @@ impl NetworkNamespace {
             }
             Ok(())
         });
-        match jh.await {
+        match res.await {
             Ok(r) => r,
             Err(e) => {
                 let err_msg = format!("Namespace removal failed: {}", e);
--- a/rtnetlink/src/ns.rs
+++ b/rtnetlink/src/ns.rs
@@ -1,8 +1,17 @@
-use nix::fcntl::OFlag;
-use nix::sched::*;
-use nix::sys::stat::Mode;
-use std::option::Option;
-use std::path::Path;
+//#[cfg(feature = "tokio")]
+use tokio::task;
+
+use crate::Error;
+use nix::{
+    fcntl::OFlag,
+    sched::*,
+    sys::{
+        stat::Mode,
+        wait::{waitpid, WaitStatus},
+    },
+    unistd::{fork, ForkResult},
+};
+use std::{option::Option, path::Path, process::exit};
 
 pub const NETNS_PATH: &str = "/var/run/netns/";
 pub const SELF_NS_PATH: &str = "/proc/self/ns/net";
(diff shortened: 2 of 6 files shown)


gabrik:
@little-dude I tested this only on one machine, so my idea was to test it across different distributions and kernel versions before marking the PR ready.

I did not notice yet any blocking on the `mount` but is it true that it runs on a separate process, I'll also do more testing from that point of view.

gabrik:
I tested on:

- Ubuntu 20.04.1 LTS - Linux 5.4.0-52-generic - x86_64 - Intel(R) Xeon(R) CPU E3-1275 v5 @ 3.60GHz
- Ubuntu 20.04.1 LTS - Linux 5.4.0-54-generic - x86_64 - Intel(R) Atom(TM) CPU  Z3740  @ 1.33GHz 
- Ubuntu 20.04.1 LTS - Linux 5.4.0-1018-raspi - aarch64 -  Raspberry Pi 4 Model B Rev 1.4
- CentOS 8.1.1911 - Linux 4.18.0-147.el8.x86_64 - x86_64 - AMD Ryzen 7 3800X 8-Core Processor 
- Gentoo - Linux 5.4.72-gentoo-x86_64 - x86_64 - AMD Ryzen 7 3700X 8-Core Processor

little-dude:
Hey @gabrik this is awesome!

I'm going to take a look tonight, and test on my machine (NixOS, kernel 5.9.8)!

gabrik:
I have done some other tests, and I managed to have it blocks, so I'll mark this ready once I'm sure everything works fine.

little-dude:
Hey @gabrik I'm a bit overwhelmed this week. I hope to review this by the end of the week, but I might not have time until this week-end, sorry.

gabrik:
@little-dude no problem, I'm still testing it, to make sure it does not block and that it works fine

little-dude (approved):
This looks great! I tested it on my machine and it seems to work :+1: 

I think it could use some more comments, and we could try split the content of this big `fork` in smaller functions, but I'd like to merge this already.

Regarding asynchrony: I think going with `block_on` is a good solution. If `mount` mount blocks the current thread, then we're goood, that's what `block_on` is for. If `mount` blocks the entire process, well then the program will hang, but at this point there's nothing we can do anyway.

rtnetlink/src/ns.rs
```diff
@@ -0,0 +1,223 @@
...
+                let err_msg = format!("Fork failed: {}", e);
+                Err(Error::NamespaceError(err_msg))
+            }
+        }
+    }
+
+    /// Remove a network namespace
+    /// This is equivalent to `ip netns del NS_NAME`.
+    pub async fn del(ns_name: String) -> Result<(), Error> {
+        let jh = task::spawn_blocking(move || {
```
little-dude, line 197: What does `jh` stand for here?

rtnetlink/src/ns.rs
```diff
@@ -0,0 +1,223 @@
...
+            let mut netns_path = String::new();
+            netns_path.push_str(NETNS_PATH);
+            netns_path.push_str(&ns_name);
+            let ns_path = Path::new(&netns_path);
+
+            if nix::mount::umount2(ns_path, nix::mount::MntFlags::MNT_DETACH).is_err() {
```
little-dude, line 203: Just to make sure I understand this correctly:

- `umount2` makes `/var/run/<namespace>` a mere directory, instead of a mount point
- `unlink` tells the system to delete `/var/run/<namespace>` after the last process that runs in that namespace exits.

rtnetlink/src/ns.rs
```diff
@@ -0,0 +1,223 @@
...
+                    Err(_) => exit(-1),
+                };
+
+                if nix::unistd::close(fd).is_err() {
+                    let _ = nix::unistd::unlink(ns_path);
+                    exit(-1)
+                }
+
+                // unshare to the new network namespace
+                if nix::sched::unshare(CloneFlags::CLONE_NEWNET).is_err() {
```
little-dude, line 149: If I understand correctly, in multithreaded environments, `unshare` has effect on the _current thread_, not the current process. So I wonder if what we're doing is correct.

rtnetlink/src/ns.rs
gabrik: It stands for job handler, I should have named it result or res. 
I'll rename it

rtnetlink/src/ns.rs
gabrik: From: https://linux.die.net/man/2/umount2
I see that:
Linux 2.1.116 added the umount2() system call, which, like umount(), unmounts a target, but allows additional flags controlling the behavior of the operation: 
And `MNT_DETACH`
MNT_DETACH (since Linux 2.4.11)
    Perform a lazy unmount: make the mount point unavailable for new accesses, and actually perform the unmount when the mount point ceases to be busy. 

Those are the same calls done by `ip netns del <ns>`, see this portion of strace:
```
...
umount2("/run/netns/ns1", MNT_DETACH)   = 0
unlink("/run/netns/ns1")                = 0
...
```

Like `ip netns del <ns>` this means that if a process is still running in the namespace the namespace is actually removed when the last process in the namespace exits, but no new process can be started on the namespace.

rtnetlink/src/ns.rs
gabrik: This was the real tricky part, indeed only the thread calling unshare does actually `unshare` itself from the parent environment, actually, I was also afraid to use the `fork` because as we are using tokio it is not really safe to fork a multithreaded process, it may hang because of a lock taken by the `parent` and that the `child` never see as released.

That said, I'm calling `unshare` in the child process, which may still be multithreaded because it starts as part of a tokio runtime.
So I do not really see other options that make this a sync interface. 
I'm really open to suggestions

gabrik:
Just to comment on the status, I found out that the system gets overwhelmed by mounts of `/run/netns` 
I'm investigating how to solve this before any possible merge

gabrik:
Fixed, I was mounting in the wrong order. Now I followed the same approach of `ip netns add <ns_name>` which is:

```
...
mount("", "/var/run/netns", 0x55bc32505175, MS_REC|MS_SHARED, NULL) = -1 EINVAL (Invalid argument)
mount("/var/run/netns", "/var/run/netns", 0x55bc32505175, MS_BIND|MS_REC, NULL) = 0
mount("", "/var/run/netns", 0x55bc32505175, MS_REC|MS_SHARED, NULL) = 0
...
```

@ffmancera can you take a look at the code?

ffmancera:
> Fixed, I was mounting in the wrong order. Now I follow the same approach of `ip netns add <ns_name>` which is:
> 
> ```
> ...
> mount("", "/var/run/netns", 0x55bc32505175, MS_REC|MS_SHARED, NULL) = -1 EINVAL (Invalid argument)
> mount("/var/run/netns", "/var/run/netns", 0x55bc32505175, MS_BIND|MS_REC, NULL) = 0
> mount("", "/var/run/netns", 0x55bc32505175, MS_REC|MS_SHARED, NULL) = 0
> ...
> ```
> 
> @ffmancera can you take a look at the code?

Sure, reviewing. Sorry for the delay.

rtnetlink/Cargo.toml
```diff
@@ -22,6 +22,7 @@ thiserror = "1"
 netlink-packet-route = { path = "../netlink-packet-route", version = "0.5" }
 netlink-proto = { path = "../netlink-proto", version = "0.4" }
 byteordered = "0.5.0"
+nix = "0.19.0"
```
ffmancera, line 25: I am not sure if this could be an issue when packaging rtnetlink on Fedora as nix seems to do not be packaged there. Introducing new dependencies can be an issue in a lot of systems due to packaging. @cathay4t

ffmancera (changes requested):
I only see one issue here.. I would need some help from @cathay4t to decide here.

rtnetlink/Cargo.toml
gabrik: Well without `nix` in order to do the whole unshare/mount stuff we will need to call directly the C code.
Anyway, I have tested it on Centos and it was working like a charm. 
What do you mean `nix` is not packaged for Fedora?

gabrik: Also just tested on a Fedora 33, it builds and works without any issue

rtnetlink/src/ns.rs
mcr: I implemented code, and I avoided starting tokio in the parent or child until the fork() and unshare(). 
Typical users will want to do something in the child (in the new namespace), so I don't think that calling fork(2) is the right thing. I think rather, all of the rest of the stuff (the mkdirs and whatnot) should be packaged into functions (not async ones), and document how to use this.

gabrik: The current code is actually mimicking the `ip netns add` which does not give the possibility to the user to execute something in the new namespace, that one is the `ip netns exec NS_NAME CMD`, in that case, I would agree with you that we should leave the possibility to the user to do something in the new namespace.

I really need this for https://github.com/eclipse-fog05/fog05/pull/208 and as I already tested across various distros and CPU architecture I would tend to just merge it.

mcr: While "ip netns add" is good to have, if we just wanted to run two scripts, we wouldn't be doing this in Rust, right? :-)
I would prefer to have all the before-fork(2), between-fork(2)-and-unshare(2), and any after-unshare(2) code refactored so that people can call it appropriately.  The result would be:

````
     NetworkNamespace::prep_for_fork(2);
     match unsafe { fork() } {
            Ok(ForkResult::Parent { child, .. }) => { NetworkNamespace::parent_processing(child); }
            Ok(ForkResult::Child) => { NetworkNamespace::child_process();  
                                                       SomeOtherProcessing(); 
                                                       NetworkNamespace::unshare_processing(); }
    }

~~~~

Will we support setns() properly?  There all sorts of interactions with tokio threads which need to be dealt with.

Comment (gabrik) [reply rtnetlink/src/ns.rs:149]: Sure I can refactor the code, but I still think that provide something like `ip netns exec NS CMD` is very difficult because only the thread calling the `unshare`+`setns`+some `mount` will be moved to the specified namespace.

Let me just explain how we use this in [fog05-linux-networking](https://github.com/eclipse-fog05/fog05/blob/master/linux-network/src/networking.rs#L926)

We have "something" asking this component to create a new network namespace, here we call the code from `netlink`, then we spawn a separate process to "manage" this new namespace (e.g. create new interfaces, nftables, etc.).

Anyway, I will refactor the code as you suggest it will also improve readability.

**Comment by cathay4t** [comment]
@little-dude The namespace support is controlled by `/proc` files, not netlink kernel interface. Why not create `nix-net-namespace` crate instead?

**Comment by cathay4t** [comment]
And namespace is not part of `rtnetlink` protocol. Why it should be listed there?

**Comment by little-dude** [comment]
It's true that these namespaces manipulations are not directly part of netlink, but there are many cases where they are used together (for instance creating veths and putting one of them in another namespace). Also, I like to see `rtnetlink` like are a library version of `iproute2`, so I think it makes sense support the equivalent of the `ip netns` commands.

Are there specific concerns about having this in `rtnetlink`? I'm not opposed to moving this to another crate, or making it an optional feature.

File: rtnetlink/Cargo.toml [hunk 5]
```diff
@@ -22,6 +22,8 @@ thiserror = "1"
 netlink-packet-route = { path = "../netlink-packet-route", version = "0.6" }
 netlink-proto = { path = "../netlink-proto", version = "0.5" }
 byteordered = "0.5.0"
+nix = "0.19.0"
+tokio = { version = "0.2.6", features = ["full"] }
```
little-dude, line 26: Sorry for missing this earlier, but as far as I can see we only need `tokio::task`, so we probably shouldn't pull in the entire tokio crate.

gabrik: I'll do a patch and fix this

cathay4t:
> Are there specific concerns about having this in `rtnetlink`? I'm not opposed to moving this to another crate, or making it an optional feature.

If `rtnetlink` is planed to mimic `iproute2`, then I am OK to have namespace support in `rtnetlink` crate.

rtnetlink/Cargo.toml
gabrik: Patch is here: https://github.com/little-dude/netlink/pull/132

rtnetlink/src/ns.rs
```diff
@@ -0,0 +1,301 @@
...
+        // Placeholder function, nothing to do here.
+        Ok(())
+    }
+
+    /// This is the parent process form the fork, it waits for the
+    /// child to exit properly
+    pub fn parent_process(child: nix::unistd::Pid) -> Result<(), Error> {
+        log::trace!("parent_process child PID: {}", child);
+        log::trace!("Waiting for child to finish...");
+        match waitpid(child, None) {
```
little-dude, line 82: I just realized that this is actually blocking for as long as the child process is alive. There's a [flag](https://docs.rs/nix/0.19.1/nix/sys/wait/struct.WaitPidFlag.html#associatedconstant.WNOHANG) to make this non-blocking but then we'd have to integrate this with tokio... The more I think about it and the more I think we might just go for the blocking version anyway. Nowadays, `/var/run` is a tmpfs in memory anyway, so `mount` and `mkdir` should be fast enough.

gabrik: Indeed, that is the part I'm really not happy about, but tested it a lot of times on different machines and it was always fast enough to not block all the process.
Indeed it should be integrated with tokio/smol or have this API as blocking.

PR #126, opened 2020-12-08
netlink-packet-route: VXLAN port attribute must be parsed as u16_be

ffmancera:
Currently it is returning a wrong value. e.g:

For a VXLAN with a dst_port "4789" the value that rust-netlink returns
is "46354".

Signed-off-by: Fernando Fernandez Mancera <ffmancera@riseup.net>


gabrik:
@ffmancera that's odd when I tested it on different machines it was fine, but I'm ok to fix it as u16_be so we are always sure about the endianness. Ideally, we should test on different CPU architecture to be 100% sure it always works

ffmancera:
> @ffmancera that's odd when I tested it on different machines it was fine, but I'm ok to fix it as u16_be so we are always sure about the endianness. Ideally, we should test on different CPU architecture to be 100% sure it always works

Yes, that would be ideal. I've found the issue on my personal machine, I would like to test this rtnetlink on ARM but needs to set everything up. Thanks!

PR #127, opened 2020-12-09
netlink-packet-route: implement VRF support

ffmancera:
VRF interfaces only has table information. This patch is introducing
InfoVrf nla.

Signed-off-by: Fernando Fernandez Mancera <ffmancera@riseup.net>


gabrik:
@ffmancera please do a run of `cargo clippy` and `cargo fmt`

ffmancera:
> @ffmancera please do a run of `cargo clippy` and `cargo fmt`

Done!

netlink-packet-route/src/rtnl/link/nlas/link_infos.rs
```diff
@@ -1345,6 +1355,54 @@ impl<'a, T: AsRef<[u8]> + ?Sized> Parseable<NlaBuffer<&'a T>> for InfoIpVlan {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum InfoVrf {
+    Unspec(Vec<u8>),
```
cathay4t, line 1360: The kernel never use `Unspec`, if that really happens, it should fit into `Other`.

cathay4t (changes requested):

netlink-packet-route/src/rtnl/link/nlas/link_infos.rs
ffmancera: I think we discussed this before. In that case, we should remove `Unspec` from the others Info* to keep consistency.

gabrik (approved):

PR #128, opened 2020-12-15
WIP: rtnetlink/addr: Support adding arbitrary IP address

cathay4t:
Introducing `AddressHandle::add_raw()` allowing user to add IP
address using netlink `AddressMesssage`.

The most useful options user could be benefit from would be:
 * `valid_lft` and `preferred_lft`. Widely used in DHCP/IPv6-RA.
 * metric

### DIFF
--- a/rtnetlink/src/addr/add.rs
+++ b/rtnetlink/src/addr/add.rs
@@ -70,6 +70,10 @@ impl AddressAddRequest {
         AddressAddRequest { handle, message }
     }
 
+    pub(crate) fn new_raw(handle: Handle, message: AddressMessage) -> Self {
+        AddressAddRequest { handle, message }
+    }
+
     /// Execute the request.
     pub async fn execute(self) -> Result<(), Error> {
         let AddressAddRequest {
--- a/rtnetlink/src/addr/handle.rs
+++ b/rtnetlink/src/addr/handle.rs
@@ -22,6 +22,10 @@ impl AddressHandle {
         AddressAddRequest::new(self.0.clone(), index, address, prefix_len)
     }
 
+    pub fn add_raw(&self, address: AddressMessage) -> AddressAddRequest {
+        AddressAddRequest::new_raw(self.0.clone(), address)
+    }
+
     /// Delete the given address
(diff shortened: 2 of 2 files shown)


cathay4t:
@little-dude The test would be slightly complex(need root, need a test veth interface). If you need it be there, I can try.
I intend to use this function in https://github.com/nispor/nispor where we will have test case for this.

ffmancera (approved):
It looks good to me. I think the best option to test this will be use it from Nispor.. Thanks!

cathay4t:
@little-dude Could you review this PR?

Meanwhile, do we have rule on when a PR can be merged? If not, how about:
 * CI should pass.
 * PR should contain test case unless state otherwise.
 * Approved by 2+ maintainer.

little-dude:
Hey @cathay4t, I agree with your rules :) But maybe for small PRs 1 approval would be enough?

rtnetlink/src/addr/add.rs
```diff
@@ -70,6 +70,10 @@ impl AddressAddRequest {
         AddressAddRequest { handle, message }
     }
 
+    pub(crate) fn new_raw(handle: Handle, message: AddressMessage) -> Self {
```
little-dude, line 73: I think it would be nice to document why someone would want to use the `_raw` methods. I'm myself not too sure about it.

little-dude: I mean I understand that you may need to tweak the messages sometimes. But in `LinkRequest` we have [`message_mut()`](https://github.com/little-dude/netlink/blob/3cf41bf7dd658abfdb07e1dd0c52bfaf5035e8f7/rtnetlink/src/link/add.rs#L298) for this. Could we use that here too, for the sake of consistency?

little-dude:
And sorry for taking so long to review this small PR @cathay4t

rtnetlink/src/addr/add.rs
cathay4t: Sure. I will use the `message_mut()` way. Thanks.

cathay4t:
Closing as exiting `message_mut()` is enough for this use case.

Issue #105, opened 2020-10-03
Migrate to tokio 0.3 / mio 0.7


little-dude:
Tokio 0.3 has been released and uses mio 0.7

little-dude:
Seems like porting `netlink-sys` is going to be easy: https://discordapp.com/channels/500028886025895936/627986131702644744/779763174437027870

We can just use [`AsyncFd`](https://docs.rs/tokio/0.3.4/tokio/io/unix/struct.AsyncFd.html) which was added in 0.3.2.

faern:
Tokio 1.0 is out and it probably only makes sense to upgrade directly to that and skip 0.3 entirely.

Issue #77, opened 2020-05-10
async_std support?


manuels:
This is a well crafted crate. Thanks for implementing it!
Is there a plan to make it work with async_std instead of tokio?

little-dude:
Hello,

That's something I'd love to have, but I have little time these days, so I'm not working on new features. However I'd gladly accept a PR for this.

Perhaps we should discuss first _how_ to do this though. Currently, there are two places where we have tokio specific code:

- `tokio.rs` in `netlink-sys`: https://github.com/little-dude/netlink/blob/master/netlink-sys/src/tokio.rs. Here we just implement `PollEvented` for the netlink socket so that we can use it with the tokio runtime. I was initially waiting for https://github.com/async-rs/async-std/issues/293 to be solved, because as far as I understand, this would automatically make the socket compatible with `async-std`. There may be a better solution for this, but I haven't used `async-std` myself, so I'm not sure.
- then there's the `netlink-proto` where we use `tokio_util::codec` in the transport layer. I think this could probably be refactored without too much effort to be framework agnostic, so it's less of a problem

joshtriplett:
As noted in in the link to `smol`, the first item isn't needed at all; because you implement `AsRawFd`, polling just works. So you'd just need a feature flag to enable `tokio` support to avoid the dependency.

The second item would require some modification to use `smol` functions rather than `tokio`, so that would have to be conditional.

gabrik:
I really need this support, I can take care of the implementation.
Where I should add this code? Inside the `netlink-sys` crate?

little-dude:
`netlink-sys` already supports `smol` since https://github.com/little-dude/netlink/pull/104.

The biggest piece is actually `netlink-proto`, which uses tokio in some places. Ideally I'd like to re-write these portions with non tokio specific bits, but it's not so trivial and I've been postponing this refactoring for a long time. I'd love some help with this and will take time to review any PR. Before starting it would be nice to discuss a plan though.

gabrik:
Sure, I'll take a look at the code to understand what the code does and I'll come back with more thoughts on this.

little-dude:
It's not too bad actually. I think we only use use `tokio_util::codec::{Decoder, Encoder};` so we could just depend on `tokio_util` with the `codec` feature. Beyond this, we also need to change this in lib.rs:

```rust
pub mod sys {
    pub use netlink_sys::{protocols, SocketAddr, TokioSocket as Socket};
}
```

[...]

PR #134, opened 2021-01-15
Upgrade tokio to 1.0

Tim-Zhang:
Fixes: #105

### DIFF
--- a/netlink-sys/Cargo.toml
+++ b/netlink-sys/Cargo.toml
@@ -9,7 +9,7 @@ keywords = ["netlink", "ip", "linux"]
 license = "MIT"
 readme = "../README.md"
 repository = "https://github.com/little-dude/netlink"
-description = "netlink sockets, with optional integration with mio and tokio"
+description = "netlink sockets, with optional integration with tokio"
 
 [dependencies]
 libc = "0.2.66"
@@ -19,16 +19,12 @@ log = "0.4.8"
 optional = true
 version = "0.3.1"
 
-[dependencies.mio]
-optional = true
-version = "0.6.21"
-
 [dependencies.tokio]
 optional = true
-version = "0.2.6"
+version = "1.0.1"
 default-features = false
 # We only depend on tokio for PollEvented
-features = ["io-driver"]
+features = ["net"]
 
 [dependencies.async-io]
 optional = true
@@ -36,7 +32,7 @@ version = "1.1"
 
 [features]
 default = []
-tokio_socket = ["mio", "tokio", "futures"]
+tokio_socket = ["tokio", "futures"]
 smol_socket = ["async-io"]
 
 
@@ -44,10 +40,10 @@ smol_socket = ["async-io"]
 netlink-packet-audit = { path = "../netlink-packet-audit" }
 
 [dev-dependencies.tokio]
-version = "0.2.6"
+version = "1.0.1"
 default-features = false
 # We only depend on tokio for PollEvented
-features = ["io-driver", "macros"]
+features = ["net", "macros", "rt-multi-thread"]
 
 [[example]]
 name = "audit_events"
--- a/netlink-sys/src/tokio.rs
+++ b/netlink-sys/src/tokio.rs
@@ -6,39 +6,12 @@ use std::{
 
 use futures::{future::poll_fn, ready};
 use log::trace;
-use mio::{event::Evented, unix::EventedFd};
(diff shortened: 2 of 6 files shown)


little-dude:
Hey, thanks a lot @Tim-Zhang I'll try to review this week-end!

little-dude:
@gabrik you might also be interested in this if you're working on https://github.com/little-dude/netlink/issues/77

mcr (commented):
I have read through the patch set, and I can't say that I exactly understand any of it: clearly I need much more knowledge of tokio.   It seems that bulk is that: Eventd trait is no more, we go from PollEvented to AsyncFd interface, and that the poll_fn() gets to make use of some refactored changes.  The comments at line 66 which go away maybe could be re-interpreted to explain how we use the guard and poll_write_ready() interface so that future maintainers can understand what's going on.

Tim-Zhang:
@mcr Thank you for your review. The reason I removed the comments at line 66 is that I think the code is clear and obviously, It's a fixed form and document in docs of `tokio`, but I overlooked that some developers are not familiar with tokio. I will add these comments back, thanks.

netlink-sys/src/tokio.rs
```diff
@@ -136,19 +100,20 @@ impl TokioSocket {
...
-        match self.0.get_ref().recv_from(buf, 0) {
-            Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {
-                trace!("poll_recv_from socket would block");
-                self.0.clear_read_ready(cx, mio::Ready::readable())?;
-                Poll::Pending
-            }
-            x => {
-                trace!("poll_recv_from {:?} bytes read", x);
-                Poll::Ready(x)
+        loop {
```
little-dude, line 103: Why do we loop here? Shoudn't we simply return `Poll::Pending` on `WouldBlock`? I'm not sure.

netlink-sys/src/tokio.rs
```diff
@@ -91,33 +61,27 @@ impl TokioSocket {
         buf: &[u8],
         addr: &SocketAddr,
     ) -> Poll<io::Result<usize>> {
-        ready!(self.0.poll_write_ready(cx))?;
-        match self.0.get_ref().send_to(buf, addr, 0) {
-            Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {
-                self.0.clear_write_ready(cx)?;
-                Poll::Pending
+        loop {
```
little-dude, line 64: Like below, I'm not sure to understand why we need to loop here.

netlink-sys/Cargo.toml
```diff
@@ -19,35 +19,31 @@ log = "0.4.8"
 optional = true
 version = "0.3.1"
 
-[dependencies.mio]
```
little-dude, line 22: Feels good to directly depend on tokio now

little-dude (commented):
Looks good to me, I just have a question about the `loop` that was added to the `poll_xxx` methods. Thanks for working on this @Tim-Zhang !

netlink-sys/src/tokio.rs
Tim-Zhang: In tokio 0.2 we only need call `clear_read_ready()` and return `Poll::Pending`, the `clear_read_ready()` will register the current task to be notified once a read readiness event is received, but there is a big change in tokio 1.0, the corresponding method `clear_ready` called by `try_io` will not register the current task we need poll the AsyncFd again through the loop.

Tim-Zhang:
@little-dude I think we should bump the major version after this pr merged

netlink-sys/src/tokio.rs
little-dude: Thanks for the explanation!

One more question: would simply clearing the readiness flag by calling [`clear_ready()`](https://docs.rs/tokio/1.0.2/tokio/io/unix/struct.AsyncFdReadyGuard.html#method.clear_ready) and returning `Poll::Pending` be enough to be notified of readiness, instead of looping?

Tim-Zhang: No, we must call `poll_read_ready()` after the `clear_ready() ` called.

The `clear_read_ready` in `PollEvented` will call the `poll_read_ready`, but the `clear_ready` in `AsyncFd` won't do that for us, we nee do it by ourselves.

little-dude:
Thanks a ton @Tim-Zhang !

PR #137, opened 2021-01-17
rtnetlink: use macros in response handling

little-dude:
This is an attempt at de-duplicating some of the code. Not sure this
makes things more readable though. Wdyt?

This adds two macros: `try_rtnl!` that returns an error if the given message is not a specific `RtnlMessage::Xxx` variant, and `try_nl!` that returns an error if the given message is a `NetlinkPayload::Error`.

### DIFF
--- a/rtnetlink/src/macros.rs
+++ b/rtnetlink/src/macros.rs
@@ -1,0 +1,29 @@
+#[macro_export]
+macro_rules! try_rtnl {
+    ($msg: expr, $message_type:path) => {{
+        use netlink_packet_route::{NetlinkMessage, NetlinkPayload, RtnlMessage};
+        use $crate::Error;
+
+        let (header, payload) = $msg.into_parts();
+        match payload {
+            NetlinkPayload::InnerMessage($message_type(msg)) => msg,
+            NetlinkPayload::Error(err) => return Err(Error::NetlinkError(err)),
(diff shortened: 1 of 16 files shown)


rtnetlink/src/macros.rs
```diff
@@ -0,0 +1,29 @@
...
+#[macro_export]
+macro_rules! try_nl {
+    ($msg: expr) => {{
+        use netlink_packet_route::NetlinkPayload;
+        use $crate::Error;
+        if let NetlinkPayload::Error(err) = $msg.payload {
+            return Err(Error::NetlinkError(err));
+        }
+    }};
+}
```
mxpv, line 29: nit: These can be even just one macro with different branches depending on the number of arguments.
Like
```rust
macro_rules! try_nl {
    ($msg: expr) => {{...}};
    ($msg: expr, $message_type:path) => {{...}};
 }
```

mxpv (approved):
This will do nicely! LGTM.

rtnetlink/src/macros.rs
little-dude: Yeah I hesitated but ultimately we're doing two slightly different things, and I thought it might be a bit confusing to have one macro for both. Not sure if others have an opinion on this.

PR #140, opened 2021-01-17
Rework rule/route builders

mxpv:
Currently netlink offers two separate rule builders (for v4 and v6), so I ended up with the code like this:
```rust
    for route in list {
            const MAIN_TABLE: u8 = packet::constants::RT_TABLE_MAIN;
            const UNICAST: u8 = packet::constants::RTN_UNICAST;
            const BOOT_PROT: u8 = packet::constants::RTPROT_BOOT;

            if is_v6 {
                // Build IP v6 request
                let mut request = self
                    .handle
                    .route()
                    .add_v6()
                    .table(MAIN_TABLE)
                    .kind(UNICAST)
                    .protocol(BOOT_PROT)
                    ...
            } else {
                // Build IP v4 request
                let mut request = self
                    .handle
                    .route()
                    .add_v4()
                    .table(MAIN_TABLE)
                    .kind(UNICAST)
                    .protocol(BOOT_PROT)
                     ...
            }
        }
```

Which is not super convenient because the majority of fields are same (except source/dest address).

This PR slightly refactors rule builder and adds a bit of rust magic, so we can have just one builder, that supports branching.
So we can do things like this:

```rust
    rule.add()
        .v4()
        .destination_prefix(dest.ip(), dest.prefix())
        .execute()
        .await?;
```

or for example above:

```rust
   let mut builder = rule.add().table(MAIN_TABLE)
                    .kind(UNICAST)
                    .protocol(BOOT_PROT)

   builder = if ip_v6 {
      builder.v6().destination_prefix(IpV6Addr).source_prefix() // v6 variation
   } else {
     builder.v4().destination_prefix(IpV4Addr).source_prefix() // v4 variation
  }

  builder.execute();

```

@little-dude would be interested in your opinion on this. 

Signed-off-by: Maksym Pavlenko <pavlenko.maksym@gmail.com>

### DIFF
--- a/rtnetlink/src/rule/add.rs
+++ b/rtnetlink/src/rule/add.rs
@@ -1,5 +1,8 @@
 use futures::stream::StreamExt;
-use std::net::{Ipv4Addr, Ipv6Addr};
+use std::{
+    marker::PhantomData,
+    net::{Ipv4Addr, Ipv6Addr},
+};
 
 use netlink_packet_route::{
     constants::*,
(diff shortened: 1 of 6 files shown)


little-dude:
Going to bed now @mxpv but I'll take a look tomorrow! Thanks!

rtnetlink/src/rule/add.rs
```diff
@@ -13,58 +16,84 @@ use netlink_packet_route::{
 use crate::{Error, Handle};
 
 /// A request to create a new rule. This is equivalent to the `ip rule add` command.
-struct RuleAddRequest {
+pub struct RuleAddRequest<T = ()> {
```
little-dude, line 19: Oh nice :+1: I didn't know that trick!

little-dude (approved):
I like it a lot. It's a nice little Rust trick, and it improves the API significantly imo :+1: 

I'll leave it open a couple days in case someone wants to look at this.

Thanks @mxpv !

mxpv:
I've applied same trick in route builder for API consistency across the project.

little-dude:
Thanks a lot @mxpv this looks great!

Tim-Zhang:
@mxpv 
```
builder = if ip_v6 {
    builder.v6().destination_prefix(IpV6Addr).source_prefix() // v6 variation
} else {
    builder.v4().destination_prefix(IpV4Addr).source_prefix() // v4 variation
}
```

It will cause an error: `if` and `else` have incompatible types, will it?

mxpv:
Correct. `v4` and `v6` return different types. What you can do is to return the future from both branches (e.g. call `execute` without await).

PR #144, opened 2021-01-22
async_std support

gabrik:
Solves #77 

I initially implemented the `AsyncStdSocket` as wrap of `SmolSocket`.

But I still think it can be used directly in the  `async_std` runtime.

Signed-off-by: gabrik <gabriele.baldoni@gmail.com>


gabrik:
I will use the `smol_socket` as a feature across the whole project for `async_std` compatibility, with default one being `tokio_socket`

I'm also adding example that uses `async_std` instead of `tokio`

gabrik:
@little-dude I see that Clippy fails because of multiple imports of `Socket`.

The imports are different and depend on the feature that was enabled, can I try by removing `--all-features` from https://github.com/little-dude/netlink/blob/master/.github/workflows/clippy-rustfmt.yml#L32 

And run Clippy two times, one for `tokio_socket` and one for `smol_socket`?

little-dude:
Hi again @gabrik :)

Thanks for the PR.

First, a disclaimer: I don't know much about async-std.

Reading the PR I don't understand why the smol socket is not enough? I thought that the `SmolSocket` implementing the [`async_io::Async`](https://docs.rs/async-io/1.3.1/async_io/struct.Async.html) trait was enough to use it with `async-std`. Is that not the case?

gabrik:
Hi @little-dude 
> Reading the PR I don't understand why the smol socket is not enough? I thought that the `SmolSocket` implementing the [`async_io::Async`](https://docs.rs/async-io/1.3.1/async_io/struct.Async.html) trait was enough to use it with `async-std`. Is that not the case?

Yes it is enough, indeed I just added some methods to match the same interface of the `TokioSocket` and then I added the `smol_socket` feature where needed. 

I tested on different machines, but `clippy` on GitHub fails, I guess because it tries with all the features at once, while `tokio_socket` and `smol_socket` should be mutually exclusive.

gabrik:
@little-dude I've split in two the Clippy run, one for the default feature set (with Tokio) and one for the `smol_socket` feature.

gabrik:
@little-dude did you had time to check this?

gabrik:
@little-dude I really need this in https://github.com/eclipse-fog05 any chance to get a review before next week?

little-dude:
Hi @gabrik,

Very sorry, I'm quite busy with work and didn't find the motivation to work on this crate lately.
I'll get to it either tonight or tomorrow.

little-dude:
This looks great, thank you! I'm merging this. I'll review a couple other PRs tonight and tomorrow, then do a release with everything if you don't mind

gabrik:
Sure, take your time to review the other PRs, I'll use master for the time being in my project, thanks again for approving this.

PR #146, opened 2021-02-04
Expose IFLA_XDP_* constants

haileysome:
It looks like these were lost in https://github.com/little-dude/netlink/commit/3ab4f0be653659d326fb0d4368ce5c688e6347f7.

I noticed these constants were left commented after the crate move. My program relies on them, can they be exposed? Thanks!


little-dude:
Oops sorry about that and thanks for the PR.

I think you'll need to expose those as either `u8`, `u16` or `u32` because  we don't expose these constants as `c_int` anymore.

```
    --> netlink-packet-route/src/rtnl/constants.rs:1147:28
     |
1147 | pub const IFLA_XDP_UNSPEC: int = 0;
     |                            ^^^ not found in this scope
```

Tim-Zhang:
@haileysome any updates?

JohnTitor (approved):

PR #147, opened 2021-02-04
Update netlink-proto to tokio 1.0, tokio-util 0.6, and bytes 1.0

charliesome:
These version bumps are all interdependent.

A summary of the changes here:

* `tokio_util::codec::Encoder` now takes the item type as a generic parameter rather than an associated type

* `BufMut::bytes_mut` has been renamed `chunk_mut` and now returns `bytes::buf::UninitSlice` rather than a slice of `MaybeUninit<u8>`. A safe API is provided to write to the slice. The semantics of the function are otherwise unchanged.

* Relaxed version constraints on tokio, tokio-util, and bytes. This will make it less likely that multiple versions of tokio are compiled into an application which will cause unpredictable runtime errors.

### DIFF
--- a/netlink-proto/Cargo.toml
+++ b/netlink-proto/Cargo.toml
@@ -12,11 +12,11 @@ repository = "https://github.com/little-dude/netlink"
 description = "async netlink protocol"
 
 [dependencies]
-bytes = "0.5.3"
+bytes = "1.0"
 log = "0.4.8"
 futures = "0.3.11"
-tokio = { version = "1.0.1", default-features = false, features = ["io-util"] }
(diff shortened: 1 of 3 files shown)


little-dude:
Hello,

Thanks a lot for the PR @charliesome and sorry for being a bit slow.
This looks good but I'll take a closer look tomorrow, since it touches potentially unsafe code, and I want to carefully check the `bytes` 1.0 doc.

Tim-Zhang (approved):
Thanks @charliesome

Tim-Zhang:
@little-dude Is there any updates?

little-dude:
Sorry for the delay. I had reviewed this already, but procrastinated. Thanks a lot @charliesome !

PR #151, opened 2021-02-21
Make Parseable<T> can parse from ?Sized type

Leo1003:
Recently, I try to implement the generic netlink protocol. But I found that I could not compile the following codes:
```rust
impl ParseableParametrized<[u8], u16> for FooMessage {
    fn parse_with_param(buf: &[u8], message_type: u16) -> Result<Self, DecodeError> {
        // Parse data...
    }
}
```
It's because that the type T isn't marked as `?Sized`. However, the parameter type pass to the function have a reference on it (`&T`). So, this should be fine. 

I also fix the doc comments of the two traits.

Note: I am not sure if this would break the backward compatibility of `netlink-packet-utils`.

### DIFF
--- a/netlink-packet-utils/src/traits.rs
+++ b/netlink-packet-utils/src/traits.rs
@@ -16,7 +16,7 @@ pub trait Emitable {
     fn emit(&self, buffer: &mut [u8]);
 }
 
-/// A `Parseable` type can be used to deserialize data into the target type `T` for which it is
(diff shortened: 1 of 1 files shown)


little-dude (approved):
Thank you very much!

> Note: I am not sure if this would break the backward compatibility of `netlink-packet-utils`.

I feel like it should not? But I can bump the minor version anyway just in case

PR #152, opened 2021-03-03
Allow to parse other audit events

hube12:
For any other event which are not audit one (but still needs to be logged AuditMessage throw a decode error which is not totally true, since a lot of events are still valid (you only take into account Audit Event but a lot of people are installed in all the events: https://github.com/linux-audit/audit-documentation/blob/main/specs/messages/message-dictionary.csv )
This PR create an Other enum field which holds all valid Audit message which are not Audit events, delete, list ...
If you don't like this kind of handling, we could make a RawAuditMessage which allow such handling

### DIFF
--- a/audit/Cargo.toml
+++ b/audit/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "audit"
-version = "0.3.0"
+version = "0.3.1"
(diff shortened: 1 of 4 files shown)


little-dude (approved):
Thanks a lot ! And sorry for the huge delay reviewing this.

PR #159, opened 2021-04-23
UnixRequest: fix protocol family

owtaylor:
The family field in was incorrectly being filled in with 0, rather than
1 (AF_UNIX). Add a test for the serialization of UnixRequest.


cathay4t (approved):
LGTM. Thanks!

PR #130, opened 2020-12-22
Add infokind and data for wireguard

benjumanji:
Add bare minimum support for creating a wireguard interface without
allocating an extra string.

PR #162, opened 2021-04-29
Grant the permission when running rtnetlink CI

JohnTitor:
A follow-up to #130.

### DIFF
--- a/.github/workflows/clippy-rustfmt.yml
+++ b/.github/workflows/clippy-rustfmt.yml
@@ -29,7 +29,7 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: clippy-tokio-socket
-        run: cargo clippy --workspace --all
+        run: cargo clippy --workspace
 
       - name: clippy-smol-socket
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -9,9 +9,6 @@ on:
 
 jobs:
   ci:
-    strategy:
-      fail-fast: true
-
```
JohnTitor, line 14: I made a mistake, this is meaningless since we don't have any matrix here. Removing it along with the CI fix.

.github/workflows/clippy-rustfmt.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: clippy-tokio-socket
-        run: cargo clippy --workspace --all
+        run: cargo clippy --workspace
```
JohnTitor, line 32: `--all` is the same as `--workspace` so it's unnecessary.

cathay4t (approved):
LGTM. Thanks for the efforts.

PR #165, opened 2021-05-31
netlink-sys: Add the SOCK_CLOEXEC for the socket

Tim-Zhang:
To prevent the socket fd from leaking to child processes.


cathay4t (approved):
LGTM. Thanks.

PR #168, opened 2021-06-06
Various netlink-proto Encoder impl fixes

stbuehler:
Afaict the missing `reserve` before line 175 might actually cause panics when serializing messages that need more space than the current buffer `capacity` provides.

Hitting `buf.remaining_mut() < msg_len` is quite unlikely, but would probably end in OOM (and loop until it does).

### DIFF
--- a/netlink-proto/src/codecs.rs
+++ b/netlink-proto/src/codecs.rs
@@ -147,12 +147,8 @@ where
 
     fn encode(&mut self, msg: NetlinkMessage<T>, buf: &mut BytesMut) -> Result<(), Self::Error> {
         let msg_len = msg.buffer_len();
-        // FIXME: we should have a max length for the buffer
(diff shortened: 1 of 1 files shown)


stbuehler:
The missing `reserve` issue can be easily demonstrated by setting `INITIAL_WRITER_CAPACITY = 0` in https://github.com/little-dude/netlink/blob/32c1f987afd5742f40fcf630e284d70d297f9cc1/netlink-proto/src/framed.rs#L145 and trying to send a message bigger than 64 bytes (the default number of bytes `BytesMut::chunk_mut()` allocates if capacity is reached). Without manipulating `INITIAL_WRITER_CAPACITY` one might have to send a message > 8k to trigger it.

JohnTitor (approved):
Looks good to me. The commit messages helped a lot to understand the relations, thanks!

PR #169, opened 2021-06-06
Cargo dep paths

stbuehler:
In my experience the best solution to link multiple crates in a single repository is using a `[patch.crates-io]` section in the virtual root manifest adding the paths to crates used by other crates in the repository.

This allows me for example to patch a single crate in the checkout into another project of mine without getting type conflicts from the other crates in the repo I didn't patch in the other project.

It also should be easier to publish crates as you don't need to hot-patch the `Crate.toml`s before release and using `--allow-dirty` or temporary commits.

A few [dev-]dependencies didn't specify a version before; I tried my best to guess something useful.

The second commit also adds the fuzz projects to the main workspace (but disables them for default build/... commands) and makes their names unique; specifying `--all` builds the fuzz projects too, so this also includes `fmt` updates in the fuzz code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,3 +11,14 @@ members = [
     "rtnetlink",
     "audit",
 ]
+
+[patch.crates-io]
+netlink-sys = { path = "netlink-sys" }
+netlink-packet-core = { path = "netlink-packet-core" }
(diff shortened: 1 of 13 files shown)


JohnTitor (approved):
Makes sense to me 👍

PR #182, opened 2021-09-21
ethtool: Add pause support of ethtool through netlink

cathay4t:
Utilizing genetlink, this patch introduce new crate for ethtool pause
support through kernel netlink interface.

Unlike other crates in this project, it is required to use `.await` for
`execute()` function due to requirement of caching family ID of generic
netlink. For example:

* rtnetlink:

```rust
    let mut links = handle.link().get().execute();
    while let Some(nl_msg) = links.try_next().await? {
    }
```

* ethtool:

```rust
    let mut pause_handle = handle.pause().get(iface_name).execute().await;
    while let Some(msg) = pause_handle.try_next().await? {
    }
```

Example code been places as `ethtool/examples/dump_pause.rs`.

### DIFF
--- a/ethtool/src/header.rs
+++ b/ethtool/src/header.rs
@@ -1,0 +1,85 @@
+use std::ffi::CString;
+
+use anyhow::Context;
+use byteorder::{ByteOrder, NativeEndian};
+use netlink_packet_utils::{
+    nla::{self, DefaultNla, NlaBuffer},
+    parsers::{parse_string, parse_u32},
+    DecodeError,
+    Parseable,
+};
+
+const ALTIFNAMSIZ: usize = 128;
+const ETHTOOL_A_HEADER_DEV_INDEX: u16 = 1;
+const ETHTOOL_A_HEADER_DEV_NAME: u16 = 2;
+const ETHTOOL_A_HEADER_FLAGS: u16 = 3;
+
+#[derive(Debug, PartialEq, Eq, Clone)]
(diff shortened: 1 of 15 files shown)


cathay4t:
@Leo1003 Since this PR is based on your `genetlink`, please take a look to make sure I understand your original design choices.

cathay4t:
@ffmancera Your review also required. The big difference on this PR is that user of this crate does not have to care about family ID any more.

ethtool/src/header.rs
```diff
@@ -0,0 +1,126 @@
+// Copyright 2021 Red Hat, Inc.
```
little-dude, line 1: Hello @cathay4t, would it be possible to not add these copyright comments? I'm ignorant of copyright law so I feel a bit uncomfortable such headers.

cathay4t: Let me remove those lines from source code even that does not changed the fact that Red Hat is owning these code implicitly.
BTW, these copyright line has been exists in kernel and many other projects for years.

cathay4t: Removed.

Leo1003: @cathay4t  You might miss `ethtool/src/header.rs`.
The copyright lines still exist in that file.

cathay4t: Oops. Thanks!

ffmancera (approved):
LGTM, thanks!

PR #185, opened 2021-09-24
ethtool: Include feature query support

cathay4t:
Equivalent to `ethtool -k` command.

Example code been places at 'ethtool/examples/dump_features.rs'
Integration test case included and enabled in CI.


ffmancera (approved):
Looks good to me! :-)

PR #191, opened 2021-10-13
Wireguard packet support

Leo1003:
This contains a draft Wireguard generic netlink packet support.
There are some issues in the current codes:
- Data converting between C struct <br/> Should this be handled by `netlink-packet-utils`?
    - `struct in_addr`
    - `struct in6_addr`
    - `struct sockaddr_in`
    - `struct sockaddr_in6`
    - `struct timespec`
- Using unsafe operations to convert between bytes slice and C struct

### DIFF
--- a/netlink-packet-wireguard/src/nlas/allowedip.rs
+++ b/netlink-packet-wireguard/src/nlas/allowedip.rs
@@ -1,21 +1,17 @@
 use crate::{
     constants::*,
-    raw::{emit_in6_addr, emit_in_addr, parse_in6_addr, parse_in_addr},
+    raw::{emit_ip, parse_ip, IPV4_LEN, IPV6_LEN},
 };
+
 use anyhow::Context;
 use byteorder::{ByteOrder, NativeEndian};
-use libc::{in6_addr, in_addr};
 use netlink_packet_utils::{
     nla::{Nla, NlaBuffer},
     parsers::*,
     traits::*,
     DecodeError,
 };
-use std::{
-    mem::{size_of, size_of_val},
-    net::IpAddr,
-};
-
+use std::{mem::size_of_val, net::IpAddr};
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum WgAllowedIpAttrs {
     Unspec(Vec<u8>),
@@ -30,8 +26,8 @@ impl Nla for WgAllowedIpAttrs {
             WgAllowedIpAttrs::Unspec(bytes) => bytes.len(),
             WgAllowedIpAttrs::Family(v) => size_of_val(v),
             WgAllowedIpAttrs::IpAddr(v) => match *v {
-                IpAddr::V4(_) => size_of::<in_addr>(),
-                IpAddr::V6(_) => size_of::<in6_addr>(),
+                IpAddr::V4(_) => IPV4_LEN,
+                IpAddr::V6(_) => IPV6_LEN,
             },
             WgAllowedIpAttrs::Cidr(v) => size_of_val(v),
         }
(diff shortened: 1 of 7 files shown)


little-dude:
Hello @Leo1003 

Apologies for not being responsive. Thanks a lot for the PR, I'll try to take a close look today.

little-dude (changes requested):
Hello,

Thanks again for your work. I'm not a wireguard user myself, so I haven't ran the example yet. 

I like the PR a lot, but I'm a bit scared of adding unsafe code, especially because I think we can avoid it here. You can see my suggested changes in your fork at https://github.com/Leo1003/netlink/pull/1

What do you think?

netlink-packet-wireguard/src/nlas/allowedip.rs
```diff
@@ -0,0 +1,80 @@
+use crate::{constants::*, raw::*};
```
little-dude, line 1: I'd prefer not to have wildcard imports for modules other than `constants`. I don't always have rust-analyzer enabled, and when I don't I like being able to quickly see from where a given symbol is being imported.

netlink-packet-wireguard/src/nlas/allowedip.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::{
+    mem::{size_of, size_of_val},
+    net::IpAddr,
+};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum WgAllowedIpAttrs {
+    Unspec(Vec<u8>),
+    Family(u16),
+    IpAddr(IpAddr),
```
little-dude, line 20: It seems that the netlink packet contains a socket address (with the port, adress family, and a couple additional fields for IPv6). So is there a reason for not making this a `SocketAddr` instead? (I've changed it to `SocketAddr` in my PR but maybe that was wrong).

Leo1003: This is the Wireguard's allowed IPs. Each entry represents a CIDR IP network that is allowed to pass through the Wireguard interface. 
It is not related to the netlink socket address.

little-dude: Ah ok. Sorry I messed this up in https://github.com/Leo1003/netlink/pull/1. I'll try to update my PR tonight.

little-dude: Done

little-dude:
@Leo1003 would you mind showing what the output of the example looks like? I don't have wireguard configured so I can't really test it.

Leo1003:
> @Leo1003 would you mind showing what the output of the example looks like? I don't have wireguard configured so I can't really test it.

```text
ListenPort: 43655
Fwmark: 0
IfIndex: 4
IfName: wg0
PrivateKey: (hidden)
PublicKey: grk00Lcpe2LwK8xkdq+Y5x/MiVtO7a7IeVd+z9MkJS8=
Peer:
  PublicKey: FfRQT9irrzS0LQ4wKP3hdcilNCPCHr9s6ccVyA1x410=
  PresharedKey: (hidden)
  LastHandshake: SystemTime { tv_sec: 1637654823, tv_nsec: 730473714 }
  PersistentKeepalive: 25
  TxBytes: 212
  RxBytes: 92
  Endpoint: <hidden here for privacy>
  AllowedIp: 10.0.3.0/24
  AllowedIp: 10.0.69.0/24
  AllowedIp: 10.96.0.0/16
```

little-dude:
Thanks for your patience @Leo1003 :) I updated my PR. Let's rebase this and get it merged!

little-dude:
I rebased and merged manually, but for some reason Github doesn't automatically mark the PR as merged.

Thanks a lot @Leo1003 ! I'll try to make a release this week.

PR #193, opened 2021-11-07
Cleanup proto visibility

stbuehler:
As preparation for #170 (and I probably need to make more PRs before #170 is ready, and I'd like to base those on this to avoid merge conflicts).

See commit message why this shouldn't be an API break.


JohnTitor (approved):
Looks fine to me.

Issue #141, opened 2021-01-17
netlink-proto: refactor


little-dude:
This is just a couple notes to myself regarding some `netlink-proto` simplifications I've been contemplating for a while:

- have a distinct codec for the audit protocol, instead of using feature flags
- don't use the tokio codec trait? They force us to copy the data in intermediate buffers, which most of the time we don't need (unless we read multiple datagrams at once, like with the audit protocol).
- have `NetlinkFramed` return a `Result` instead of an `Option`
- rename `NetlinkFramed` into `Transport` maybe?
- don't use unbounded channels? But then we have to handle errors when they get full. Not sure how to do that

Also:

- https://github.com/little-dude/netlink/issues/138
- https://github.com/little-dude/netlink/issues/139

little-dude:
Also noticed that we need to upgrade to newer `tokio-utils` and `bytes` versions.

PR #195, opened 2021-11-13
Refactor base types

stbuehler:
Hi again.

#170 currently is broken because the `workaround-audit-bug` feature does bad things (it basically ignores the message length, which means the payload suddenly also contains the padding in my "test payload with `length % 4 != 0`" tests)

So I tried to get rid of that "feature" (also mentioned in #141 as "todo item"), and added `Codec` as type parameter to `Connection`.

As this is a major change anyway I tried to cleanup a few other things; those 4 commits should mostly be unrelated (I recommend reviewing them individually), so the subject of this PR just says "refactor" - but they all are API changes, so it makes somewhat sense to combine them (imho).

If this gets accepted I'm gonna basically have to redo #170 - so I'd like to get some feedback before I'll go back to #170 :)


stbuehler:
Hm. Seems I got some issues to fix:
- `tokio` vs `smol`: seems the tests found a case where neither one is selected. Actually those features conflict too, which is not supported by cargo - probably should get fixed (in the same "bump" cycle?): it either needs another type parameter or internal boxing, and probably different entry points
- The examples want to use `Vec<u8>` for recv... perhaps use `bytes::buf::UninitSlice` instead of `&mut [MaybeUninit<u8>]` - that would add `bytes` dependency to sys.

netlink-sys/src/smol.rs
```diff
@@ -66,14 +66,14 @@ impl SmolSocket {
             trace!("poll_recv_from socket is ready for reading");
 
             match self.0.get_ref().recv_from(buf, 0) {
-                Ok(x) => {
-                    trace!("poll_recv_from {:?} bytes read", x);
-                    return Poll::Ready(Ok(x));
-                }
-                Err(_would_block) => {
+                Err(e) if e.kind() == io::ErrorKind::WouldBlock => {
```
little-dude: oops, thanks. Maybe we should still distinguish between `Err` and `Ok` below though, otherwise the trace message might be slightly confusing when there's an error.

netlink-packet-core/src/traits.rs
```diff
@@ -1,16 +1,15 @@
 use crate::NetlinkHeader;
 use std::error::Error;
 
-/// A `NetlinkDeserializable` type can be used to deserialize a buffer
-/// into the target type `T` for which it is implemented.
-pub trait NetlinkDeserializable<T> {
```
little-dude, line 6: Really like this change, thanks!

netlink-packet-audit/src/codec.rs
```diff
@@ -0,0 +1,135 @@
...
+use bytes::BytesMut;
+use netlink_packet_core::{
+    NetlinkBuffer,
+    NetlinkDeserializable,
+    NetlinkMessage,
+    NetlinkSerializable,
+};
+pub(crate) use netlink_proto::{NetlinkCodec, NetlinkMessageCodec};
+
+/// audit specific implemenation of [`NetlinkMessageCodec`] due to the
```
little-dude, line 12: s/implemenation/implementation

netlink-packet-audit/src/codec.rs
```diff
@@ -0,0 +1,135 @@
...
+/// Among the known bugs in kernel audit messages:
+/// - `nlmsg_len` sometimes contains the padding too (it shouldn't)
+/// - `nlmsg_len` sometimes doesn't contain the header (it really should)
+///
+/// See also:
+/// - https://blog.des.no/2020/08/netlink-auditing-and-counting-bytes/
+/// - https://github.com/torvalds/linux/blob/b5013d084e03e82ceeab4db8ae8ceeaebe76b0eb/kernel/audit.c#L2386
+/// - https://github.com/mozilla/libaudit-go/issues/24
+/// - https://github.com/linux-audit/audit-userspace/issues/78
+pub struct NetlinkAuditCodec {
```
little-dude, line 24: I think this can just be

```
pub struct NetlinkAuditCodec;
```

netlink-proto/src/codecs.rs
```diff
@@ -7,36 +7,48 @@ use netlink_packet_core::{
...
-    phantom: PhantomData<T>,
-}
+/// Protocol to serialize and deserialize messages to and from datagrams
+///
+/// This is separate from `tokio_util::codec::{Decoder, Encoder}` as the implementations
+/// rely on the buffer containing full datagrams; they won't work well with simple
+/// bytestreams.
+///
+/// Officially there should be exactly one implementation of this, but the audit
+/// subsystem ignores way to many rules of the protocol, so they need a separate
```
little-dude, line 18: s/to/too

netlink-proto/src/framed.rs
```diff
@@ -1,45 +1,50 @@
...
+    codecs::NetlinkMessageCodec,
+    sys::{Socket, SocketAddr},
+};
+use netlink_packet_core::{NetlinkDeserializable, NetlinkMessage, NetlinkSerializable};
 
-pub struct NetlinkFramed<C> {
+pub struct NetlinkFramed<T, C> {
     socket: Socket,
-    codec: C,
+    msg_type: PhantomData<fn(T) -> T>, // invariant
```
little-dude, line 21: Do you mind explaining why you chose `fn(T) -> T` instead of just `T`, and what you mean by the "invariant" comment?

netlink-sys/src/socket.rs
```diff
@@ -252,34 +255,50 @@ impl Socket {
...
+        //             +------+-------+   +--------+-------+
+        //            /                \ /                  \
+        let buf_ptr = chunk.as_mut_ptr() as *mut libc::c_void;
+        let buf_len = chunk.len() as libc::size_t;
 
         let res = unsafe { libc::recvfrom(self.0, buf_ptr, buf_len, flags, addr_ptr, addrlen_ptr) };
         if res < 0 {
             return Err(Error::last_os_error());
+        } else {
+            let written = std::cmp::min(buf_len, res as usize);
```
little-dude, line 275: Interesting, I would have assumed `res <= buf_len` is always true. Did you add this "just in case", or is my assumption wrong?

EDIT: the answer if a couple lines below: my assumption is wrong.

little-dude (approved):
Wow this is great!

The `Codec` thing has been bothering me from the start, and I didn't know how to improve the situation. I also love the `AsyncSocket` trait. I'm slightly bothered by making the `netlink-sys` API rely on `bytes`  but the improvement is really worth it. Also, thank you for the soundness fix.

I'm going to let this open one day or two in case someone else wants to take a look, but I love it!

netlink-sys/src/smol.rs
stbuehler: Afaict this was and is the same for both the tokio and smol code; I'd recommend to change this some other time. (Also this code is replaced in the `AsyncSocket` commit, but with similar trace logging.)

netlink-packet-audit/src/codec.rs
stbuehler: It would expose the type "constructor", without actually needing an instance of it ever.

But if you prefer a union type I can certainly change that (and probably for `NetlinkCodec` too?).

netlink-proto/src/framed.rs
stbuehler: https://doc.rust-lang.org/nomicon/phantom-data.html has a few words on the matter.

I think invariant seems a good choice until someone comes up with a proper usecase for using covariant or contravariant (which, afaict, is mostly about lifetimes). `*mut T` would be invariant too but wouldn't derive `Send` nor `Sync` - `fn(T) -> T` derives `Send` and `Sync` afaict.

I'll add a comment in the code for it.

netlink-sys/src/socket.rs
stbuehler: I'll simply copy the comment; also `MSG_TRUNC` is enough for this, `MSG_PEEK` is not needed afaict - I will change the comment accordingly.

netlink-packet-audit/src/codec.rs
stbuehler: should be fixed now

netlink-proto/src/codecs.rs
stbuehler: should be fixed now

netlink-packet-audit/src/codec.rs
little-dude: Ah I see, this is great let's keep it as is.

genetlink/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "genetlink"
-version = "0.1.0"
+version = "0.2.0"  # TODO: drop this comment - already bumped version for trait changes
```
little-dude, line 3: I think you forgot to remove those comments?

little-dude:
Thanks @stbuehler

Do you want me to make new releases after this is merged, or do you want to rebase #170 and have releases once we have batch messages support?

genetlink/Cargo.toml
stbuehler: I wanted to make sure the affected packages get "major" bumps; the idea was that the comments get removed when actually releasing (instead of doing minor bumps).

If you prefer doing it a different way just let me know.

little-dude: Oh I see. That's fine then I'll remove them when releasing.

stbuehler:
> I'm slightly bothered by making the `netlink-sys` API rely on `bytes`  but the improvement is really worth it.

Wanted to write a few words about that: `&mut [MaybeUninit]` can't be used without `unsafe` (on the "user" side - the implementation calling libc is unsafe anyway obviously). I think there are plans to come up with something better in `std` itself (for the `Read` trait), and perhaps then we can move to that.

> Do you want me to make new releases after this is merged, or do you want to rebase #170 and have releases once we have batch messages support?

I don't need a release to continue working on #170 - I'd say that is up to you :)

little-dude:
> Wanted to write a few words about that: `&mut [MaybeUninit]` can't be used without `unsafe` (on the "user" side - the implementation calling libc is unsafe anyway obviously). I think there are plans to come up with something better in `std` itself (for the `Read` trait), and perhaps then we can move to that.

Sounds good. If you have links to specific RFCs/PRs or issues I'm interested!

For now, let's get this in! I'll make a release after you've rebased #170 

Thanks for the quality work :)

PR #197, opened 2021-11-21
license: Include SPDX-License-Identifier: MIT to source file

cathay4t:
This command is used to generate the changes:

    find -type f -name \*.rs -exec sed -i \
        '1s/^/\/\/ SPDX-License-Identifier: MIT\n\n/' {} \;

Adding license header to each source file is suggested by Fedora where
the `licensecheck` command is used for license vetting.


little-dude (approved):

JohnTitor:
So, you want to distribute the crates on Fedora, right? I think we should also add a check on CI not to forget this header by mistake.

PR #199, opened 2021-11-22
netlink-packet-route: re-introduce rich NLAs

little-dude:
The Netlink API is not stable, and the packet format may change from
one kernel version to another or even from a CPU architecture to
another. In an ideal world, we'd have conditional compilation based on
the kernel version and/or system architecture. But at the moment, we
just follow whatever is the current format at the time of
implementation.

This led to several issues, which prompted a PR that simply removes
the problematic NLAs:
https://github.com/little-dude/netlink/pull/51

However, not parsing these NLAs means allocating a buffer and do the parsing a
posteriori which is wasteful and inconvenient. Therefore, this commit
re-introduces these problematic NLAs but behind a feature flag.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,6 +46,7 @@ jobs:
         run: |
           cd netlink-packet-route
           cargo test
+          cargo test --features rich_nlas
 
       - name: test (netlink-packet-sock-diag)
         run: |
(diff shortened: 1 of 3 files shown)


little-dude:
cc @pinkisemils 

This should not affect you though because this is opt-in.

pinkisemils:
I did try building our app with these changes, but there were some dependency issues. However, the changes look good in so far as they won't break anything for us. Thank you very much!

PR #202, opened 2021-11-24
rtnetlink: add replace option to Add*Request

Tuetuopay:
This option sets the `NLM_F_REPLACE` instead of the `NLM_F_EXCL` flag, allowing "create or update" patterns.

My current usecase is for managing neighbors, for which I have a control plane, but it can happen that the neighbor is learnt from the dataplane first, and controlplane second. In this case, I want to set the neighbor as `NOARP` to prevent it from expiring. The easiest is to allow the kernel to just replace the neighbor with the one provided rather than create-fail-delete-create.

### DIFF
--- a/rtnetlink/src/addr/add.rs
+++ b/rtnetlink/src/addr/add.rs
@@ -13,6 +13,7 @@ use netlink_packet_route::{
     NLM_F_ACK,
     NLM_F_CREATE,
     NLM_F_EXCL,
+    NLM_F_REPLACE,
     NLM_F_REQUEST,
 };
 
@@ -22,6 +23,7 @@ use crate::{try_nl, Error, Handle};
 pub struct AddressAddRequest {
     handle: Handle,
     message: AddressMessage,
+    replace: bool,
 }
 
 impl AddressAddRequest {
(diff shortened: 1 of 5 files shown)


rtnetlink/src/addr/add.rs
```diff
@@ -68,17 +70,28 @@ impl AddressAddRequest {
...
-        AddressAddRequest { handle, message }
+        AddressAddRequest {
+            handle,
+            message,
+            replace: false,
+        }
+    }
+
+    /// Replace existing matching address.
+    pub fn replace(self, replace: bool) -> Self {
```
little-dude, line 81: I don't know if we really need a boolean here. Are there cases where we'll call this with `false`? It seems to me that we'll either do `.replace(true)` or not call this method at all.

Tuetuopay: Yep you're right, I tend to leave too much flexibility... Changing this in all `Add*Request`.

little-dude (approved):
Thanks, this is great :+1: 

Just like #203 I'm going to wait for a day before merging in case someone else wants to review this.

PR #203, opened 2021-11-24
rtnetlink/neigh: add new_bridge helper to add AF_BRIDGE neighs

Tuetuopay:
This allows to manipulate the FDB in addition to the ARP table.

While the same could be accomplished with `.message_mut()`, this is less cumbersome:

Before:
```rust
// 0.0.0.0 as a dummy IP, we don't need it for generic FDB entries
let mut req = handle.neighbours()
    .add(index, [0, 0, 0, 0].into())
    .state(NUD_PERMANENT)
    .link_local_address(&[0x8c, 0x8c, 0xde, 0xad, 0xbe, 0xef]);
let mut msg = req.message_mut();
msg.header.family = AF_BRIDGE as u8;
req.execute().await?;
```

After:
```rust
handle.neighbours().add_bridge(index, &[0x8c, 0x8c, 0xde, 0xad, 0xbe, 0xef]).execute().await?;
```

### DIFF
--- a/rtnetlink/src/neighbour/add.rs
+++ b/rtnetlink/src/neighbour/add.rs
@@ -40,6 +40,19 @@ impl NeighbourAddRequest {
         NeighbourAddRequest { handle, message }
     }
 
+    pub(crate) fn new_bridge(handle: Handle, index: u32, lla: &[u8]) -> Self {
+        let mut message = NeighbourMessage::default();
+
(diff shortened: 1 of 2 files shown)


little-dude (approved):
Thanks, looks neat. I'll wait for tomorrow before merging in case someone else wants to take a look.

PR #204, opened 2021-11-25
netlink-sys: reduce lines within unsafe blocks

mcginty:
As a reader looking at the unsafe blocks in this project, I feel like it helps to constrict unsafe blocks to exactly the lines that *require* they be within an unsafe block, and to not include lines that are actually safe.

So personally this version is more clear to me than wrapping an entire function in a big `unsafe { }` which makes me have to figure out which lines are the particularly high-risk ones.

Thanks so much for this project! I've been using it in [innernet](https://github.com/tonarino/innernet) and am looking forward to hopefully having more to contribute here down the line (excited to see there's already a wireguard netlink PR open).


little-dude (approved):
Hello,

Thanks for the PR! I wrote this code a long time ago, and I'm not sure why I decided to have such big unsafe blocks at the time. I agree that this version is clearer.

I didn't know about innernet, it looks very neat! I'm glad to see that these netlink crates are useful.

PR #208, opened 2021-11-29
rtnl/link: change add_name_filter to match_name

Tuetuopay:
Since kernel commit [a3d1289126e7b][1], it is possible to set the `IFLA_IFNAME` attribute in a `RTM_GETLINK` message, allowing to get a link by its name without dumping the whole link list.

In large setups, this speeds up massively the library, going from "painfully slow" to "really quick" (in my application with thousands of links, a lookup by name went from hundreds of ms to hundreds of µs).

This is based off the equivalent function in [libnl][2].

[1]: https://github.com/torvalds/linux/commit/a3d1289126e7b14307074b76bf1677015ea5036f
[2]: https://github.com/thom311/libnl/blob/master/lib/route/link.c#L1376

### DIFF
--- a/rtnetlink/examples/add_address.rs
+++ b/rtnetlink/examples/add_address.rs
@@ -33,7 +33,7 @@ async fn add_address(link_name: &str, ip: IpNetwork, handle: Handle) -> Result<(
     let mut links = handle
         .link()
         .get()
-        .set_name_filter(link_name.to_string())
+        .match_name(link_name.to_string())
         .execute();
     if let Some(link) = links.try_next().await? {
         handle
--- a/rtnetlink/examples/add_neighbour.rs
+++ b/rtnetlink/examples/add_neighbour.rs
@@ -31,7 +31,7 @@ async fn add_neighbour(link_name: &str, ip: IpAddr, handle: Handle) -> Result<()
     let mut links = handle
         .link()
         .get()
-        .set_name_filter(link_name.to_string())
+        .match_name(link_name.to_string())
         .execute();
     if let Some(link) = links.try_next().await? {
         handle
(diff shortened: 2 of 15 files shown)


little-dude:
This is great! I was not really satisfied with the filtering thing.

Since we're on this, would you mind adding some doc on the name filter to explain that if users only want to filter on the name, they should rather use `match_name`?

Also out of curiosity, is the application you mentioned public?

Tuetuopay:
Well since I removed the old `set_name_filter`, there's no more doc to add :D Though I'll add the same doc as libnl, telling this requires a recent-ish kernel (2009 vintage) for the option to work. Maybe you'd like to keep the function for precisely those people?

And sadly not, the application is not public, I'm doing this for work. Though I think I can talk a bit about it: it's managing lots (thousands) of VXLAN networks across several netns in a cloud provider environment.

little-dude:
Ah yes sorry. Maybe we can just remove the [`FilterBuilder`](https://github.com/little-dude/netlink/blob/5836d7357d75f94d59cac1f81f408697a567920e/rtnetlink/src/link/get.rs#L95) entirely then.

> it's managing lots (thousands) of VXLAN networks across several netns in a cloud provider environment.
Merge state

Sounds cool! The network namespaces are a dark corner for me.

Tuetuopay:
@little-dude removed the whole `LinkFilterBuilder` then, with the appropriate warning on `match_name`, regarding kernel version.

> Sounds cool! The network namespaces are a dark corner for me.

Yeah control plane stuff is fun! And namespaces are too, and very very useful. Just takes a little while to get used to...

little-dude:
This is great, thank you!

mcr:
> Well since I removed the old `set_name_filter`, there's no more doc to add :D Though I'll add the same doc as libnl, telling this requires a recent-ish kernel (2009 vintage) for the option to work. Maybe you'd like to keep the function for precisely those people?

I would have prefered that set_name_filter() was deprecated before being removed.
This was rather puzzling to solve.

Issue #139, opened 2021-01-17
netlink-proto: handle `ENOBUFS` errors


little-dude:
To quote the man(7):

> However, reliable transmissions from kernel to user are impossible in any case. The kernel can't send a netlink message if the socket buffer is full: the message will be dropped and the kernel and the user-space process will no longer have the same view of kernel state. It is up to the application to detect when this happens (via the ENOBUFS error returned by recvmsg(2)) and resynchronize. 

Currently `NetlinkFramed` will just log the error and continue, not letting the user handle the error

idanski:
I actually encountered this with audit recently! 
So I'm very interested in this feature, was thinking of suggesting someway to handle it but didn't got around to it.

Let me know if I can help somehow. :pray:

little-dude:
Oh, sorry that you ran into it. I think at the moment, the only way around it is not to use `netlink-proto` to be honest. I'm not exactly sure how I'm going to tackle this yet, because this issue also relates to the refactoring I mentioned in https://docs.rs/tokio/1.0.2/tokio/io/unix/struct.AsyncFdReadyGuard.html#method.clear_ready

Would you mind explaining in more detail what you're doing to trigger this? And if you were able to catch the ENOBUFS errors, how would you handle them, exactly?

idanski:
sure no problem.

recreating this is not easy but I ran audit in multicast mode + enabled other products that listen on auditd.
on some machines the multicast socket sometimes received ENOBUFS and the stream was indeed consumed.

To test I forked your repo and ignored ENOBUFS, printing a log instead and I saw it was working well + logs were printed.
I thought maybe the correct way of handling it was either to pass error codes to ignore to `NetlinkFramed` or somehow return the error message from the stream (maybe as a wrapped `Result<Message, NetlinkError>`) I didn't really looked into it yet.

Again thanks for the awesome crate and let me know if I can help in any way.

Tuetuopay:
I so happen to be hitting this too, I'm working on working around it. If you need some way to reproduce it, it's very easy to trigger it by adding a lot of neighbours to a link, and deleting it:

```rust
use futures::stream::{StreamExt, TryStreamExt};
use rtnetlink::packet::{constants::NUD_INCOMPLETE, link::nlas::Nla};
use rtnetlink::{constants::RTMGRP_NEIGH, new_connection, sys::SocketAddr};
use std::net::Ipv4Addr;

#[tokio::main(flavor = "current_thread")]
async fn main() {
[...]

PR #210, opened 2021-12-04
Add flag attribute support to WireGuard NLAs

mcginty:
These flag attributes are used when setting device properties.


little-dude (approved):
Thanks a lot :) I'm excited to see these crates to be used in bigger projects. However, I don't know if you're using `netlink-proto` in innernet, but if you do, please be aware of this issue: https://github.com/little-dude/netlink/issues/139

This is the issue I'd like to fix in priority but I don't have that much time to dedicate to it.

little-dude:
Ah I just looked at your innernet PR, looks like you're not using netlink-proto so you're good :)

PR #214, opened 2021-12-27
add documentation on macvlan function's mode argument

mcr:
I had to grep around the place to determine what the mode argument was.
There is probably a better way to present this information here.


little-dude:
Thanks @mcr  !

PR #216, opened 2021-12-28
Don't depend on async-std

dvc94ch:
before:
```
cargo tree --no-default-features --features smol_socket --edges no-dev | wc -l
169
```
after:
```
cargo tree --no-default-features --features smol_socket --edges no-dev | wc -l
135
```

### DIFF
--- a/rtnetlink/Cargo.toml
+++ b/rtnetlink/Cargo.toml
@@ -15,7 +15,7 @@ description = "manipulate linux networking resources via netlink"
 test_as_root = []
 default = ["tokio_socket"]
 tokio_socket = ["netlink-proto/tokio_socket", "tokio"]
-smol_socket = ["netlink-proto/smol_socket", "async-std"]
+smol_socket = ["netlink-proto/smol_socket", "async-global-executor"]
 
 [dependencies]
 futures = "0.3.11"
@@ -25,7 +25,7 @@ netlink-packet-route = "0.9"
 netlink-proto = { default-features = false, version = "0.8" }
 nix = "0.22.0"
 tokio = { version = "1.0.1", features = ["rt"], optional = true}
-async-std = { version = "1.9.0", features = ["unstable"], optional = true}
+async-global-executor = { version = "2.0.2", optional = true }
 
 [dev-dependencies]
 env_logger = "0.8.2"
--- a/rtnetlink/src/ns.rs
+++ b/rtnetlink/src/ns.rs
@@ -20,7 +20,7 @@ where
     F: FnOnce() -> R + Send + 'static,
     R: Send + 'static,
 {
-    async_std::task::spawn_blocking(fut).await
+    async_global_executor::spawn_blocking(fut).await
 }
 
 // only tokio enabled, so use tokio


little-dude:
Hello,

Thanks for the PR. I haven't been following the latest async ecosystem evolutions but I assume `async-std` has been split into multiple crates, and `async-global-allocator` is one of them?

dvc94ch:
It's nothing new, it happened when async-std adopted async-io. for most libraries I'd recommend using async-io and async-global-executor. In fact I've never needed anything else.

dvc94ch:
just in case it wasn't clear, yes async-std uses async-global-executor to spawn tasks.

JohnTitor:
Why did you add 8cd6a4b98db56599c7dc5145e9e7202488ce71a8? The `patch.crates-io` section on root Cargo.toml should be sufficient.

dvc94ch:
hadn't seen that. but I haven't seen patch.crates-io used like that before. the normal way do depend on a crate in the same repo is setting the path. I noticed the paths were missing when I patched rtnetlink in my crate. if the paths aren't set correctly it fails to build because the crates are not actually compatible with the ones from crates.io.

dvc94ch:
`[patch.crates-io]` is useful when referring to code outside of the repo.

JohnTitor:
> hadn't seen that. but I haven't seen patch.crates-io used like that before. the normal way do depend on a crate in the same repo is setting the path. 

It's _your_ normal way, actually [actix-web do the same way](https://github.com/actix/actix-web/blob/2b2de298001c7718952a7e8647f6b1a2736c7259/Cargo.toml#L133-L143).
Setting paths to each netlink crate is annoying, and we don't have to care about it when using `patch.crates-io`. This repo contains many crates so it makes much sense here.
I would NOT accept these commits, please drop them.
(Moreover, this PR is for async-std and any unrelated changes should go to another PR.)

little-dude:
> It's nothing new, it happened when async-std adopted async-io. for most libraries I'd recommend using async-io and async-global-executor. In fact I've never needed anything else.

Thanks!

As @JohnTitor said the `patch.crates-io` trick was done on purpose so we'd prefer not to change this.

rtnetlink/src/lib.rs
```diff
@@ -40,6 +40,7 @@ pub mod constants;
 pub use netlink_packet_route as packet;
 pub mod proto {
     pub use netlink_proto::{Connection, ConnectionHandle, Error, ErrorKind};
+    pub use netlink_proto::packet::NetlinkMessage;
```
little-dude: I'm not against adding re-exports but could this be done in a separate PR?

dvc94ch:
Well looks like actix is your crate. It's not my way, I explained that crates fail to build with this method as the patch.crates-io is not for declaring dependencies. Here's an example: https://github.com/rust-lang/rust/blob/master/library/std/Cargo.toml

little-dude:
> Well looks like actix is your crate. It's not my way, I explained that crates fail to build with this method as the patch.crates-io is not for declaring dependencies. Here's an example: [rust-lang/rust@`master`/library/std/Cargo.toml](https://github.com/rust-lang/rust/blob/master/library/std/Cargo.toml?rgh-link-date=2021-12-29T11%3A28%3A32Z)

I personally don't have a strong opinion on this but since this particular change is controversial, I suggest backing it out from this PR, and opening an issue where we can discuss the pros and cons of this approach. For reference here is the PR where it was implemented: https://github.com/little-dude/netlink/pull/169

little-dude:
>  I noticed the paths were missing when I patched rtnetlink in my crate.

So is this a blocking issue for you? If so, would making a release unblock you for now?

dvc94ch:
Like your strategy. Yes, getting a release would make this argument uninteresting enough for me to move on. However the next contributor might run into the same issue. Removed the commits and opened a new PR.

little-dude:
> Like your strategy.

I just want you to have a short term solution while we're agreeing on a longer term solution :)

> Yes, getting a release would make this argument uninteresting enough for me to move on.

Cool, I'm a work, I'll make a release tonight if you can wait for a couple hours.

> However the next contributor might run into the same issue.

This seems problematic indeed. 

> Removed the commits and opened a new PR.

Thank you!

dvc94ch:
The same problem happens if you use `git = "https://github.com/little-dude/netlink"`

little-dude:
New release is available @dvc94ch : https://github.com/little-dude/netlink/releases/tag/20211229

dvc94ch:
Thanks

PR #218, opened 2021-12-28
the example "ip link" command had a typo

mcr:
if NAME is considered the variable, then the order should be "name NAME" rather than "NAME name"


little-dude:
Thank you!

PR #219, opened 2021-12-29
create_macvlan example should use symbolic constants

mcr:
The example should show how to get the right constants, and then use them to validate that they are correct.


mcr:
Confused. Seems that it is already done.

PR #213, opened 2021-12-13
bridge: add support for setting `IFLA_BRIDGE_VLAN_INFO`

flouthoc:
Extends `InfoBridge` to support Flags
    
Add support for setting `IFLA_BRIDGE_VLAN_INFO` to bridge management nested attributes.
    
            BRIDGE_VLAN_INFO_MASTER = 1
            BRIDGE_VLAN_INFO_PVID = 4
            BRIDGE_VLAN_INFO_UNTAGGED = 8
            BRIDGE_VLAN_INFO_RANGE_BEGIN = 16
            BRIDGE_VLAN_INFO_RANGE_END = 32

PR #212, opened 2021-12-13
Bridge: add support for specifying `IFLA_BRIDGE_FLAGS` and implement `Flags`

flouthoc:
Extends `InfoBridge` to support `Flags`

Add support for setting `IFLA_BRIDGE_FLAGS` to bridge management nested attributes. Can accept flags such `BRIDGE_FLAGS_MASTER` and `BRIDGE_FLAGS_SELF`

PR #222, opened 2022-01-03
nlas: Add type for `AfSpecBridge` which allows `links` to configure various bridge flags with `SETLINK` calls

flouthoc:
Following PR allows `handle.link().set(` or `RTM_SETLINK` calls to configure various `bridge` attributes like.

* `IFLA_BRIDGE_FLAGS`
* `IFLA_BRIDGE_VLAN_INFO`

With this PR its easy to implement functionality like
```console
bridge vlan add ...
```

This PR also corrects some changes made in 
https://github.com/little-dude/netlink/pull/212 and https://github.com/little-dude/netlink/pull/213

### DIFF
--- a/netlink-packet-route/src/rtnl/link/nlas/af_spec_bridge.rs
+++ b/netlink-packet-route/src/rtnl/link/nlas/af_spec_bridge.rs
@@ -1,5 +1,4 @@
 // SPDX-License-Identifier: MIT
-
 use anyhow::Context;
 
 use crate::{
--- a/netlink-packet-route/src/rtnl/link/nlas/af_spec_bridge.rs
+++ b/netlink-packet-route/src/rtnl/link/nlas/af_spec_bridge.rs
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: MIT
+
+use anyhow::Context;
+
+use crate::{
+    constants::*,
+    nlas::{self, DefaultNla, NlaBuffer},
+    parsers::parse_u16,
+    traits::Parseable,
+    DecodeError,
+};
+
+use byteorder::{ByteOrder, NativeEndian};
+
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub enum AfSpecBridge {
+    Flags(u16),
+    VlanInfo(Vec<u8>),
+    Other(DefaultNla),
+}
+
+impl nlas::Nla for AfSpecBridge {
+    fn value_len(&self) -> usize {
+        use self::AfSpecBridge::*;
+        match *self {
+            VlanInfo(ref bytes) => bytes.len(),
+            Flags(_) => 2,
+            Other(ref nla) => nla.value_len(),
+        }
+    }
+
+    fn emit_value(&self, buffer: &mut [u8]) {
+        use self::AfSpecBridge::*;
+        match *self {
+            Flags(value) => NativeEndian::write_u16(buffer, value),
+            VlanInfo(ref bytes) => {
+                (&mut buffer[..bytes.len()]).copy_from_slice(bytes.as_slice());
+            }
+            Other(ref nla) => nla.emit_value(buffer),
+        }
+    }
+
+    fn kind(&self) -> u16 {
+        use self::AfSpecBridge::*;
+        match *self {
+            Flags(_) => IFLA_BRIDGE_FLAGS,
+            VlanInfo(_) => IFLA_BRIDGE_VLAN_INFO,
+            Other(ref nla) => nla.kind(),
+        }
+    }
+}
+
+impl<'a, T: AsRef<[u8]> + ?Sized> Parseable<NlaBuffer<&'a T>> for AfSpecBridge {
+    fn parse(buf: &NlaBuffer<&'a T>) -> Result<Self, DecodeError> {
+        use self::AfSpecBridge::*;
+
+        let payload = buf.value();
+        Ok(match buf.kind() {
+            IFLA_BRIDGE_VLAN_INFO => VlanInfo(payload.to_vec()),
+            IFLA_BRIDGE_FLAGS => {
+                Flags(parse_u16(payload).context("invalid IFLA_BRIDGE_FLAGS value")?)
+            }
+            kind => Other(DefaultNla::parse(buf).context(format!("Unknown NLA type {}", kind))?),
+        })
+    }
+}
--- a/netlink-packet-route/src/rtnl/link/nlas/mod.rs
+++ b/netlink-packet-route/src/rtnl/link/nlas/mod.rs
@@ -9,6 +9,9 @@ pub use self::inet6::*;
 mod af_spec_inet;
 pub use self::af_spec_inet::*;
 
+mod af_spec_bridge;
+pub use self::af_spec_bridge::*;
+
 mod link_infos;
 pub use self::link_infos::*;
 
@@ -142,8 +145,8 @@ pub enum Nla {
     Map(Vec<u8>),
     // AF_SPEC (the type of af_spec depends on the interface family of the message)
     AfSpecInet(Vec<AfSpecInet>),
-    // AfSpecBridge(Vec<AfSpecBridgeNla>),
-    AfSpecBridge(Vec<u8>),
+    AfSpecBridge(Vec<AfSpecBridge>),
+    //AfSpecBridge(Vec<u8>),
     AfSpecUnknown(Vec<u8>),
     Other(DefaultNla),
 }
@@ -177,7 +180,6 @@ impl nlas::Nla for Nla {
                 | Broadcast(ref bytes)
                 | PermAddress(ref bytes)
                 | AfSpecUnknown(ref bytes)
-                | AfSpecBridge(ref bytes)
                 | Map(ref bytes)
                 | ProtoDownReason(ref bytes)
                 => bytes.len(),
@@ -223,7 +225,7 @@ impl nlas::Nla for Nla {
             Info(ref nlas) => nlas.as_slice().buffer_len(),
             PropList(ref nlas) => nlas.as_slice().buffer_len(),
             AfSpecInet(ref nlas) => nlas.as_slice().buffer_len(),
-            // AfSpecBridge(ref nlas) => nlas.as_slice().buffer_len(),
+            AfSpecBridge(ref nlas) => nlas.as_slice().buffer_len(),
             Other(ref attr)  => attr.value_len(),
         }
     }
@@ -258,7 +260,6 @@ impl nlas::Nla for Nla {
                 | Broadcast(ref bytes)
                 | PermAddress(ref bytes)
                 | AfSpecUnknown(ref bytes)
-                | AfSpecBridge(ref bytes)
                 | Stats(ref bytes)
                 | Stats64(ref bytes)
                 | Map(ref bytes)
@@ -309,7 +310,7 @@ impl nlas::Nla for Nla {
             Info(ref nlas) => nlas.as_slice().emit(buffer),
             PropList(ref nlas) => nlas.as_slice().emit(buffer),
             AfSpecInet(ref nlas) => nlas.as_slice().emit(buffer),
-            // AfSpecBridge(ref nlas) => nlas.as_slice().emit(buffer),
+            AfSpecBridge(ref nlas) => nlas.as_slice().emit(buffer),
             // default nlas
             Other(ref attr) => attr.emit_value(buffer),
         }
@@ -500,7 +501,15 @@ impl<'a, T: AsRef<[u8]> + ?Sized> ParseableParametrized<NlaBuffer<&'a T>, u16> f
                     }
                     AfSpecInet(nlas)
                 }
-                AF_BRIDGE => AfSpecBridge(payload.to_vec()),
+                AF_BRIDGE => {
+                    let mut nlas = vec![];
+                    let err = "invalid IFLA_AF_BRIDGE value";
+                    for nla in NlasIterator::new(payload) {
+                        let nla = nla.context(err)?;
+                        nlas.push(af_spec_bridge::AfSpecBridge::parse(&nla).context(err)?);
+                    }
+                    AfSpecBridge(nlas)
+                }
                 _ => AfSpecUnknown(payload.to_vec()),
             },
             IFLA_LINKINFO => {
--- a/netlink-packet-route/src/rtnl/constants.rs
+++ b/netlink-packet-route/src/rtnl/constants.rs
@@ -395,11 +395,11 @@ pub const IFLA_INFO_XSTATS: u16 = 3;
 pub const IFLA_INFO_SLAVE_KIND: u16 = 4;
 pub const IFLA_INFO_SLAVE_DATA: u16 = 5;
 // Bridge flags
-pub const IFLA_BRIDGE_FLAGS: u16 = 47;
+pub const IFLA_BRIDGE_FLAGS: u16 = 0;
 pub const BRIDGE_FLAGS_MASTER: u16 = 1; /* Bridge command to/from master */
 pub const BRIDGE_FLAGS_SELF: u16 = 2; /* Bridge command to/from lowerdev */
 
-pub const IFLA_BRIDGE_VLAN_INFO: u16 = 48;
+pub const IFLA_BRIDGE_VLAN_INFO: u16 = 2;
 pub const BRIDGE_VLAN_INFO_MASTER: u16 = 1;
 pub const BRIDGE_VLAN_INFO_PVID: u16 = 4;
 pub const BRIDGE_VLAN_INFO_UNTAGGED: u16 = 8;
--- a/netlink-packet-route/src/rtnl/link/nlas/link_infos.rs
+++ b/netlink-packet-route/src/rtnl/link/nlas/link_infos.rs
@@ -805,7 +805,6 @@ pub enum InfoBridge {
     // FIXME: what type is this? putting Vec<u8> for now but it might
     // be a boolean actually
     FdbFlush(Vec<u8>),
-    Flags(u16),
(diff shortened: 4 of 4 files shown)


flouthoc:
@little-dude @cathay4t PTAL. Let me know if any example file is needed with this, however simple example like this works.

```rust
    /* bridge vlan add ... */
    let mut links = handle.link().get().match_name("my-bridge-1".to_string().clone()).execute();
    if let Some(link) = links.try_next().await? {

        let mut request = handle.link().set(link.header.index);
        request.message_mut().header.interface_family = AF_BRIDGE as u8;
        request.message_mut().header.flags = 0;
        let my_struct = BridgeVlanInfo { id: 6u16, data: 100u16};
        let bytes: &[u8] = unsafe { any_as_u8_slice(&my_struct) };

        let bridge_flags = AfSpecBridge::Flags(BRIDGE_FLAGS_SELF);
        let bridge_vlan_info = AfSpecBridge::VlanInfo(bytes.to_vec());
        request.message_mut().nlas.push(Nla::AfSpecBridge(vec![bridge_flags, bridge_vlan_info]));
        request.execute().await.map_err(|e| format!("{}", e));
    } else {
        println!("no link link found");
    }
    Ok(())

**Comment by flouthoc** [comment]
@cathay4t @little-dude Gentle reminder. Were you guys able to take a look at this. No rush.

**Comment by flouthoc** [comment]
Even if this is merged I would be not able to use quickly due to version dependency across all netlink crates will have to wait for release of `netlink-packet-route` and that has to be updated in `rtnetlink` master.

So no rush here.

**Comment by flouthoc** [comment]
> Even if this is merged I would be not able to use quickly due to version dependency across all netlink crates will have to wait for release of `netlink-packet-route` and that has to be updated in `rtnetlink` master.
> 
> So no rush here.

Not related to this PR but we could avoid this drift it if all crates share same version and are released together. But it has its own problems. I'll open a separate GH issue to discuss this.

**Comment by little-dude** [comment]
Hey @flouthoc,

Just a heads up, I'm not ignoring you. I've had covid symptoms since last Friday. It's getting better but I'm not going to look at pending PRs for a couple more days.

**Comment by flouthoc** [comment]
> Hey @flouthoc,
> 
> Just a heads up, I'm not ignoring you. I've had covid symptoms since last Friday. It's getting better but I'm not going to look at pending PRs for a couple more days.

Ah Please take care. This PR is not urgent we can do this later. Get well soon.

File: netlink-packet-route/src/rtnl/link/nlas/af_spec_bridge.rs [hunk 1]
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: MIT
+use anyhow::Context;
+
+use crate::{
```
little-dude, line 4: Some symbols are not being used:

```
warning: unused imports: `Emitable`, `NlasIterator`
 --> netlink-packet-route/src/rtnl/link/nlas/af_spec_bridge.rs:6:41
  |
6 |     nlas::{self, DefaultNla, NlaBuffer, NlasIterator},
  |                                         ^^^^^^^^^^^^
7 |     parsers::parse_u16,
8 |     traits::{Emitable, Parseable},
  |              ^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `BigEndian`
  --> netlink-packet-route/src/rtnl/link/nlas/af_spec_bridge.rs:12:17
   |
12 | use byteorder::{BigEndian, ByteOrder, NativeEndian};
   |                 ^^^^^^^^^
```

netlink-packet-route/src/rtnl/link/nlas/af_spec_bridge.rs
```diff
@@ -0,0 +1,67 @@
...
+
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub enum AfSpecBridge {
+    Flags(u16),
+    VlanInfo(Vec<u8>),
+    Other(DefaultNla),
+}
+
+impl nlas::Nla for AfSpecBridge {
+    #[rustfmt::skip]
```
little-dude, line 22: I don't think we need to skip rustfmt here. The formatted output should be the same.

netlink-packet-route/src/rtnl/link/nlas/af_spec_bridge.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn value_len(&self) -> usize {
+        use self::AfSpecBridge::*;
+        match *self {
+            VlanInfo(ref bytes) => bytes.len(),
+            Flags(_) => 2,
+            Other(ref nla) => nla.value_len(),
+        }
+    }
+
+    #[rustfmt::skip]
```
little-dude, line 32: Same as above, no need to skip rustfmt

netlink-packet-route/src/rtnl/link/nlas/mod.rs
```diff
@@ -500,7 +501,15 @@ impl<'a, T: AsRef<[u8]> + ?Sized> ParseableParametrized<NlaBuffer<&'a T>, u16> f
                     }
                     AfSpecInet(nlas)
                 }
-                AF_BRIDGE => AfSpecBridge(payload.to_vec()),
+                AF_BRIDGE => {
+                    let mut nlas = vec![];
+                    let err = "invalid IFLA_AF_SPEC value";
```
little-dude, line 506: This should be `"invalid AF_BRIDGE value"`.

little-dude (changes requested):

little-dude:
> Even if this is merged I would be not able to use quickly due to version dependency across all netlink crates will have to wait for release of `netlink-packet-route` and that has to be updated in `rtnetlink` master.
> 
> So no rush here.

I can make a release once this is merged. I'll try to make more smaller releases as we go. Big releases are a pain.

little-dude:
> @little-dude @cathay4t PTAL. Let me know if any example file is needed with this, however simple example like this works.
> 
> ```rust
>     /* bridge vlan add ... */
>     let mut links = handle.link().get().match_name("my-bridge-1".to_string().clone()).execute();
>     if let Some(link) = links.try_next().await? {
> 
>         let mut request = handle.link().set(link.header.index);
>         request.message_mut().header.interface_family = AF_BRIDGE as u8;
>         request.message_mut().header.flags = 0;
>         let my_struct = BridgeVlanInfo { id: 6u16, data: 100u16};
>         let bytes: &[u8] = unsafe { any_as_u8_slice(&my_struct) };
> 
>         let bridge_flags = AfSpecBridge::Flags(BRIDGE_FLAGS_SELF);
>         let bridge_vlan_info = AfSpecBridge::VlanInfo(bytes.to_vec());
>         request.message_mut().nlas.push(Nla::AfSpecBridge(vec![bridge_flags, bridge_vlan_info]));
>         request.execute().await.map_err(|e| format!("{}", e));
>     } else {
>         println!("no link link found");
>     }
>     Ok(())
> ```

Probably no need for an example in `netlink-packet-route` but an integration with `rtnetlink` and an example there would be awesome :)

flouthoc:
@little-dude thanks for the review. I'll fix nits and add a example by this weekend.

flouthoc:
@little-dude Sorry I got stuck in few other things. But I need this PR so i'll get back to this, and would probably close this in this week. So if not in this release we can get this in by next release. 

So please don't stop release because of this but I'll make sure that it will not miss this in another release.

netlink-packet-route/src/rtnl/link/nlas/mod.rs
flouthoc: Fixed.

flouthoc:
@little-dude I didn't add example yet but other than that resolved all the nits you requested.

stbuehler:
I'm a little worried that `IFLA_AF_SPEC` seems to be used with two different protocols.

* one is described in [uapi/linux/if_bridge.h](https://github.com/torvalds/linux/blob/4ea3c6425269d33da53c79d539ce9554117cf4d4/include/uapi/linux/if_bridge.h#L111-L128) and contains the bridge values "flat".
* the second is described in [uapi/linux/if_link.h](https://github.com/torvalds/linux/blob/4ea3c6425269d33da53c79d539ce9554117cf4d4/include/uapi/linux/if_link.h#L273-L289) and describes it as a map from address family (as "attribute type") to (nested) af specific data (for `AF_BRIDGE` probably the same as the first one, just nested one level deeper? not sure.)

I think the first one is only to be used if `rtgen_family == AF_BRIDGE` (I'm guessing this becomes `interface_family` in the parser), and the second one for everything else. The current implementation seems completely broken; especially `emit` would need to consider the current family too.

flouthoc:
@stbuehler Yeah but issue is i'm not able to use `AfSpecBridge` at all, do you have any suggestions for `emit` block ?

stbuehler:
Most attributes (and `IFLA_AF_SPEC` too) only can occur once, but the nested content forms a list/map again. So you probably should start with something like this:

```rust
pub enum AfSpecElem {
    AfSpecIpv4(AfSpecInet),
    AfSpecIpv6(AfSpecInet),
    AfSpecBridge(AfSpecBridge),
    AfSpecUnknown {
        family: u8,
        data: Vec<u8>,
    }
}

/// rtnl::link::nlas::Nla
pub enum Nla {
    ...
    AfSpec(Vec<AfSpecElem>),
}
```

(I'm not a fan of these `Vec<SomeNla>` things and would rather use something like this:
```rust
pub struct AfSpec {
    ipv4: Option<AfSpecInet>,
    ipv6: Option<AfSpecInet>,
    bridge: Option<AfSpecBridge>,
    unknown: Vec<(u8, Vec<u8>)>,
}

/// rtnl::link::nlas::Nla
pub enum Nla {
    ...
    AfSpec(AfSpec),
}
```

But that is not how it is done in this library, and more work to implement the encode/decode things for.)

When parsing you could now map the `rtgen_family == AF_BRIDGE` case by parsing an inner `AfSpecBridge` and just put it into the nested data as it would have been an inner nested entry. Then you'll have to undo that on `emit` by somehow passing the `rtgen_family` - the code for emitting nested Nlas is not that complex, so you should be able to duplicate it with local functions (instead of traits, or build a generic trait variant similar to the parser case).

Or you could add a different `Nla` variant for the bridge family:

```rust
/// rtnl::link::nlas::Nla
pub enum Nla {
    ...
    // only with `rtgen_family != AF_BRIDGE`
    AfSpec(Vec<AfSpecElem>),
    // only with `rtgen_family == AF_BRIDGE`
    AfSpecBridge(AfSpecBridge),
}
```

On parsing you should have the family as context to distinguish the two cases, and on emit it would be the users responsibility to choose the correct one.

(Note that I'm still not sure whether the contained `AfSpecBridge` data is actually the same in both cases; you should probably check the kernel for that.)

stbuehler:
Hm. It seems I didn't look closely enough at `AfSpecInet`. It actually already has the nested parts for the inner mapping (although it is quite generic for lots of families, and only three are actually implemented in the kernel afaict: `AF_INET`, `AF_INET6`, `AF_MCTP`. `AF_BRIDGE` isn't used here.). I think the name is a little bit misleading though :)

This `AfSpecInet` format is (again afaict) used for messages for all families but `AF_BRIDGE`; in this case the bridge specific format is used.

So this does look alright after all.

flouthoc:
> Hm. It seems I didn't look closely enough at `AfSpecInet`. It actually already has the nested parts for the inner mapping (although it is quite generic for lots of families, and only three are actually implemented in the kernel afaict: `AF_INET`, `AF_INET6`, `AF_MCTP`. `AF_BRIDGE` isn't used here.). I think the name is a little bit misleading though :)
> 
> This `AfSpecInet` format is (again afaict) used for messages for all families but `AF_BRIDGE`; in this case the bridge specific format is used.
> 
> So this does look alright after all.

@stbuehler Thank you so much for spending time on this and confirming. I really appreciate it :smile: . Just waiting for reviewers to take final look @cathay4t @little-dude PTAL

netlink-packet-route/src/rtnl/link/nlas/mod.rs
```diff
@@ -500,7 +501,15 @@ impl<'a, T: AsRef<[u8]> + ?Sized> ParseableParametrized<NlaBuffer<&'a T>, u16> f
...
                 }
-                AF_BRIDGE => AfSpecBridge(payload.to_vec()),
+                AF_BRIDGE => {
+                    let mut nlas = vec![];
+                    let err = "invalid IFLA_AF_BRIDGE value";
+                    for nla in NlasIterator::new(payload) {
+                        let nla = nla.context(err)?;
+                        nlas.push(af_spec_bridge::AfSpecBridge::parse(&nla).context(err)?);
+                    }
+                    AfSpecBridge(nlas)
```
cathay4t, line 511: The process might be wrong. The kernel code indicate a interface could have mulitple family's `IFLA_AF_SPEC`. Let me do some digging.

cathay4t (commented):
I need more time on kernel code reading to decide whether current code is correct or not.

cathay4t:
> `IFLA_AF_SPEC`


The kernel code `net/core/rtnetlink.c` function `rtnl_fill_link_af` explained `IFLA_AF_SPEC` is two layer nested netlink message, like:

```
[ AF_BRIDGE: [ AfSpecBridge::VlanInfo],
  AF_INET: [ AfSpecInet::IpmrTableId] ]
```

This is the snippet of kernel code, in case you would like to  read C:

```c
	af_spec = nla_nest_start_noflag(skb, IFLA_AF_SPEC);
	if (!af_spec)
		return -EMSGSIZE;

	list_for_each_entry_rcu(af_ops, &rtnl_af_ops, list) {
		struct nlattr *af;
		int err;

		if (!af_ops->fill_link_af)
			continue;

		af = nla_nest_start_noflag(skb, af_ops->family);
```

Just some notes on kernel code reading.

cathay4t:
This is the partial output of `sudo strace -s 1024 ip netconf show dev wifi0`, indicating the `IFLA_AF_SPEC` is two layers nested NLA.

But, apparently , this is too complex for this PR to fix. I intend to merge this PR and continue the fix in new PR for this `IFLA_AF_SPEC` stuff and also finish off the VLAN filtering of bridge.

```
[{nla_len=408, nla_type=IFLA_AF_SPEC},
    [
        [{nla_len=140, nla_type=AF_INET}, [
            {nla_len=136, nla_type=IFLA_INET_CONF}, [
                [IPV4_DEVCONF_FORWARDING-1] = 0,
                <omitted>
                [IPV4_DEVCONF_ARP_EVICT_NOCARRIER-1] = 1]]],
        [{nla_len=264, nla_type=AF_INET6}, [
            [{nla_len=8, nla_type=IFLA_INET6_FLAGS}, IF_READY],
            [{nla_len=20, nla_type=IFLA_INET6_CACHEINFO},
                {
                    max_reasm_len=65535,
                    tstamp=3794,
                    reachable_time=37584,
                    retrans_time=1000}],
            [{nla_len=232, nla_type=IFLA_INET6_CONF},
                [[DEVCONF_FORWARDING] = 0,
                <omitted>
                [DEVCONF_NDISC_EVICT_NOCARRIER] = 1]]]]]]
```

cathay4t (approved):
Looks good to me. Will continue the work in another PR.

flouthoc:
Thanks @cathay4t for the review and findings, let me know if there is anything i can help with.

cathay4t:
Hi @flouthoc , we can merge this once we fix https://github.com/little-dude/netlink/pull/222/files#r858245460

netlink-packet-route/src/rtnl/link/nlas/mod.rs
cathay4t: My fault. The kernel has `rtnl_register(PF_BRIDGE, RTM_GETLINK, NULL, rtnl_bridge_getlink, 0);` which indicate kernel is not using two layer nested NLA for `PF_BRIDGE`

cathay4t:
Conclusion on this `IFLA_AF_SPEC`:
 * For `PF_UNSPEC`/`AF_UNSPEC`, the `IFLA_AF_SPEC` is __two__ layer nested NLA as mentioned https://github.com/little-dude/netlink/pull/222#issuecomment-1109304841 , The `br_af_ops` does not register the `fill_link_af` function, hence bridge interface will not append vlan filtering information in this query or dump.
 * For `PF_BRIDGE`, the `IFLA_AF_SPEC` is __single__ layer nested NLA for `IFLA_BRIDGE_VLAN_INFO`, `IFLA_BRIDGE_MRP` and etc. And this is only available for `NLM_F_DUMP`.
 
For kernel code reference: `rtnetlink_init()` of `/net/core/rtnetlink.c` has:

```c
	rtnl_register(PF_UNSPEC, RTM_GETLINK, rtnl_getlink,
		      rtnl_dump_ifinfo, 0);
        // omitted
	rtnl_register(PF_BRIDGE, RTM_GETLINK, NULL, rtnl_bridge_getlink, 0);
```

netlink-packet-route/src/rtnl/link/nlas/mod.rs
```diff
@@ -500,7 +501,15 @@ impl<'a, T: AsRef<[u8]> + ?Sized> ParseableParametrized<NlaBuffer<&'a T>, u16> f
                     }
                     AfSpecInet(nlas)
                 }
-                AF_BRIDGE => AfSpecBridge(payload.to_vec()),
+                AF_BRIDGE => {
+                    let mut nlas = vec![];
+                    let err = "invalid IFLA_AF_BRIDGE value";
```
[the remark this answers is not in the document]
cathay4t: Amended

cathay4t:
I have amended the incorrect error message mentioned in https://github.com/little-dude/netlink/pull/222#discussion_r858245460 and merged. Thanks for the contribution!

flouthoc:
@cathay4t My bad sorry for the delay and thanks for merging this :)

PR #225, opened 2022-01-05
netlink-packet-wireguard: various serialization correctness fixes

### LINKED ISSUE #224: `netlink-packet-wireguard`: SetDevice serialization issues
While GetDevice and parsing works great, SetDevice encounters `Invalid argument (os error 22)` whenever the `WgDeviceAttrs::Peers` attribute is present.

It seems like there's probably some discrepancy in the encoding of the peer list - it generates an unparseable netlink message:

Example code:
```rust
    let genlmsg: GenlMessage<Wireguard> = GenlMessage::from_payload(Wireguard {
        cmd: WireguardCmd::SetDevice,
        nlas: vec![
            WgDeviceAttrs::IfName(argv[1].clone()),
            WgDeviceAttrs::PrivateKey([10u8; 32]),
            WgDeviceAttrs::Peers(vec![vec![WgPeerAttrs::PublicKey([1u8; 32]), WgPeerAttrs::PersistentKeepalive(25)]])
        ],
    });
    let mut nlmsg = NetlinkMessage::from(genlmsg);
    nlmsg.header.flags = NLM_F_REQUEST | NLM_F_ACK;

    nlmsg.finalize();
    let mut buf = [0; 4096];
    nlmsg.serialize(&mut buf);
    let len = nlmsg.buffer_len();

    // panics
    let response = NetlinkMessage::<GenlMessage<Wireguard>>::deserialize(&buf[..len]).unwrap();
```

mcginty:
Fixes #224

I used the very useful `nlmon` functionality to capture the netlink traffic from the official `wg` tool, and this is what an example NLA buffer with peer attributes looks like.

It seems that we were missing a special "nest" NLA to mark array beginning and ends. I don't know if there's a more generic solution for the rest of the netlink crates - if this is true for all serializations of slices/arrays - but for now this fixes the WireGuard problem.

Example debug output from the "official" `wg` netlink code:
```
parsing buffer as WgDeviceAttrs (len: 44, kind: 8, nested: true, byteorder: false) NlaBuffer { buffer: [44, 0, 8, 128, 40, 0, 0, 128, 36, 0, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1] }
outer NLA(len: 40, kind: 0, nested: true, byteorder: false) NlaBuffer { buffer: [40, 0, 0, 128, 36, 0, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1] }
```

I didn't want to break any other code so I kept this scoped to the WireGuard package, but I'm happy to make a more universal change if there's a safe way.


netlink-packet-wireguard/src/nlas/mod.rs
```diff
@@ -5,3 +5,25 @@ mod peer;
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct NestedSlice<'a, T: Nla>(&'a [T]);
+impl<'a, T: Nla> Nla for NestedSlice<'a, T> {
+    fn value_len(&self) -> usize {
+        self.0.buffer_len()
+    }
+
+    fn kind(&self) -> u16 {
+        0
```
little-dude, line 19: If I understand correctly, `NestedSlice` is intended to be used in various places. We can't know for sure that the NLA type will be 0 everytime, can we?

little-dude (commented):
Thanks @mcginty ! This looks good, I just have one question about `NestedSlice`. Also there's some formatting to fix.

little-dude:
> It seems that we were missing a special "nest" NLA to mark array beginning and ends. I don't know if there's a more generic solution for the rest of the netlink crates - if this is true for all serializations of slices/arrays - but for now this fixes the WireGuard problem.

This seems to be specific to wireguard. The "generic solution" in the netlink crates for nested NLAs has been so far to implement `Nla::is_nested` method like you did.

mcginty:
I updated this PR to instead use an explicit `WgPeer` and `WgAllowedIp` type instead of `NestedSlice` since you're right, all we know is that the `type` of each of those two is `0`, not generically. I think this also is cleaner for users of the API since they can get the correct `buffer_len()` for each peer this way.

mcginty:
@little-dude when you have time, would you mind releasing a new version of `netlink-packet-wireguard` with these fixes so that I can in turn release my `wireguard-control` crate to crates.io? thanks again for all your work.

little-dude:
Done @mcginty 

If you don't mind I also added you as an author in the Cargo.toml file.

mcginty:
Thanks @little-dude!!

PR #229, opened 2022-01-22
utils: Optimized calculation align

wllenyj:
Use bitwise operations instead of division.
The length value must be less than `usize::MAX - 4`, otherwise it will
panic. The result is the same as the original logic.

Signed-off-by: wllenyj <wllenyj@linux.alibaba.com>

### DIFF
--- a/netlink-packet-utils/src/nla.rs
+++ b/netlink-packet-utils/src/nla.rs
@@ -19,6 +19,17 @@ pub const NLA_F_NESTED: u16 = 0x8000;
 pub const NLA_F_NET_BYTEORDER: u16 = 0x4000;
 /// Identify the bits that represent the type of a netlink attribute.
 pub const NLA_TYPE_MASK: u16 = !(NLA_F_NET_BYTEORDER | NLA_F_NESTED);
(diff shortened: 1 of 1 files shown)


wllenyj:
@little-dude   Have some comments for the pr? 
Thanks.

wllenyj:
@little-dude   Have some comments for the pr? 
Thanks.

little-dude (changes requested):
Thanks for this improvement and for adding tests :) This looks great, but could we just rename `RTA` into `NLA` for consistency? I know in the kernel it's called `RTA` and perhaps it was the wrong decision to diverge from that in the first place, but now that we went for `NLA`, I'd rather stay consistent.

wllenyj:
> Thanks for this improvement and for adding tests :) This looks great, but could we just rename `RTA` into `NLA` for consistency? I know in the kernel it's called `RTA` and perhaps it was the wrong decision to diverge from that in the first place, but now that we went for `NLA`, I'd rather stay consistent.

Sure, fixed.
Thank you for review.

little-dude (approved):
Thanks!

PR #232, opened 2022-01-28
Add traffic control support

wllenyj:
Currently only `u32 filter` and `qdisc ingress` is supported.


wllenyj:
@little-dude  @cathay4t  Do you have any comments on this pr?
Thanks.

wllenyj:
Will all tc Kind have the same name problem? If not, there is no need to do qdisc/filter/class distinction in TcMessage, which would simplify the implementation.

But I found that, for example, `tc qdisc` has `htb`, and `tc class` also has `htb`, but they can share the same attr without conflict.

So I will try to implement it by removing the TcMessage template class.

netlink-packet-route/src/rtnl/tc/message.rs
```diff
@@ -1,31 +1,69 @@
...
+impl<A> Default for TcMessage<A> {
+    fn default() -> Self {
+        Self {
+            nlas: Vec::new(),
+            ..Default::default()
+        }
+    }
+}
+
+impl<A> From<i32> for TcMessage<A> {
```
little-dude, line 43: I feel like this `From` impl is not needed. We can simply do:
```rust
let mut tc_msg = TcMessage<Class>::default();
tc_msg.header.index = 1;
```
Also it's not obvious what the relation between an i32 and a TcMessage is. One has to look at the code to realize it corresponds to the index.

netlink-packet-route/src/rtnl/tc/nlas/qdisc/ingress.rs
```diff
@@ -0,0 +1,15 @@
...
+
+/// Ingress qdisc
+/// 
+/// This effectively allows you to police incoming traffic, before it even enters
+/// the IP stack.
+/// The ingress qdisc itself does not require any parameters. It differs from
+/// other qdiscs in that it does not occupy the root of a device. 
+
+#[derive(Debug, PartialEq, Eq, Clone, Default)]
+pub struct Ingress();
```
little-dude, line 11: Out of curiosity what is the difference between:

```rust
pub struct Ingress;
```
and
```rust
pub struct Ingress();
```

netlink-packet-route/src/rtnl/tc/nlas/class/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    fn emit_value(&self, buffer: &mut [u8]) {
+        match self {
+            Self::Other(o) => buffer.copy_from_slice(o.as_slice()),
+        }
+    }
+
+    fn kind(&self) -> u16 {
+        unreachable!("the parent nla will return TCA_OPTIONS")
```
little-dude, line 30: I'd rather not panic in a publicly available method.

If I understand correctly, `Class` is not an NLA, but rather the payload of an `TCA_OPTIONS` NLA. In that case, I think it shouldn't implement the `Nla` trait (and same goes for `Filter` and `Qdisc`. Or am I mis-understanding something?

little-dude:
Thanks a lot for the hard work @wllenyj. Generally speaking I like the approach you've taken. I need to spend a bit more time reviewing this though because I'm not familar with `tc`. I already left a couple questions for you.

wllenyj:
> Thanks a lot for the hard work @wllenyj. Generally speaking I like the approach you've taken. I need to spend a bit more time reviewing this though because I'm not familar with `tc`. I already left a couple questions for you.

I have re-implemented a version and removed the template parameter of TcMessage.
I'll get right back to finishing the patch.

netlink-packet-route/src/rtnl/tc/nlas/qdisc/ingress.rs
wllenyj: Reimplemented this piece.

netlink-packet-route/src/rtnl/tc/nlas/class/mod.rs
wllenyj: The `TcOpt` is used for different message type distribution. Just use `parse_with_param` and pass in the `kind` parameter to distinguish.

wllenyj:
The ci failure may be an environmental issue that works when my environment is running.

```
---- traffic_control::add_filter::test::test_new_filter stdout ----
thread 'traffic_control::add_filter::test::test_new_filter' panicked at 'assertion failed: `(left == right)`
  left: `4`,
 right: `5`', rtnetlink/src/traffic_control/add_filter.rs:299:21
```

netlink-packet-route/src/rtnl/tc/constants.rs
```diff
@@ -0,0 +1,97 @@
...
+
+/// Action attributes
+pub const TCA_ACT_UNSPEC: u16 = 0;
+pub const TCA_ACT_KIND: u16 = 1;
+pub const TCA_ACT_OPTIONS: u16 = 2;
+pub const TCA_ACT_INDEX: u16 = 3;
+pub const TCA_ACT_STATS: u16 = 4;
+pub const TCA_ACT_PAD: u16 = 5;
+pub const TCA_ACT_COOKIE: u16 = 6;
+pub const __TCA_ACT_MAX: u16 = 7;
```
little-dude, line 55: I don't think we need this, since we define `TCA_ACT_MAX` just below.

netlink-packet-route/src/rtnl/tc/constants.rs
```diff
@@ -0,0 +1,97 @@
...
+
+pub const TCA_ACT_TAB: u16 = 1; // TCA_ROOT_TAB
+pub const TCAA_MAX: u16 = 1;
+
+/// Mirred action attr
+pub const TCA_MIRRED_UNSPEC: u16 = 0;
+pub const TCA_MIRRED_TM: u16 = 1;
+pub const TCA_MIRRED_PARMS: u16 = 2;
+pub const TCA_MIRRED_PAD: u16 = 3;
+pub const __TCA_MIRRED_MAX: u16 = 4;
```
little-dude, line 91: I'm confused by this: why not just define `TCP_MIRRED_MAX = 3`?

netlink-packet-route/src/rtnl/tc/message.rs
```diff
@@ -26,6 +33,30 @@ impl TcMessage {
     }
 }
 
+impl From<i32> for TcMessage {
```
little-dude, line 36: I'd prefer to have a method for this, it would be more explicit. `From` and `Into` are conversion traits, and we're not really doing a conversion, we're creating an object and initializing one of it's attributes to a certain value.

```rust
impl TcMessage {
    /// Create a new `TcMessage` with the given index
    fn with_index(index: i32) -> Self {
        // ...
    }
}
```

netlink-packet-route/src/rtnl/tc/message.rs
```diff
@@ -60,7 +91,9 @@ impl Emitable for TcMessage {
 
     fn emit(&self, buffer: &mut [u8]) {
         self.header.emit(buffer);
-        self.nlas.as_slice().emit(buffer);
+        self.nlas
+            .as_slice()
+            .emit(&mut buffer[self.header.buffer_len()..]);
```
little-dude, line 96: Geez... Thanks you for fixing this.

netlink-packet-route/src/rtnl/tc/nlas/action/mirred.rs
```diff
@@ -0,0 +1,122 @@
...
+            TCA_MIRRED_TM => Tm(payload.to_vec()),
+            TCA_MIRRED_PARMS => Parms(TcMirred::parse(&TcMirredBuffer::new_checked(payload)?)?),
+            _ => Other(DefaultNla::parse(buf)?),
+        })
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Default)]
+pub struct TcMirred {
+    // TcGen
```
little-dude, line 71: What does this comment mean? Is it just a leftover?

netlink-packet-route/src/rtnl/tc/nlas/filter/mod.rs
```diff
@@ -0,0 +1,3 @@
+// SPDX-License-Identifier: MIT
+
+pub mod u32;
```
little-dude, line 3: I got confused for a second :laughing:

netlink-packet-route/src/rtnl/tc/nlas/filter/u32.rs
```diff
@@ -0,0 +1,240 @@
+// SPDX-License-Identifier: MIT
+
+/// U32 filter
```
little-dude, line 3: Thank you for adding some documentation btw :)

netlink-packet-route/src/rtnl/tc/nlas/qdisc/prio.rs
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+
+//#[derive(Debug, PartialEq, Eq, Clone)]
```
little-dude, line 3: Should this file go? Or are you planning to add more?

rtnetlink/src/traffic_control/add_filter.rs
```diff
@@ -0,0 +1,334 @@
...
+                    ))]),
+                ],
+            }]),
+        ];
+        self.u32(u32_nla)
+    }
+}
+
+#[cfg(test)]
+mod test {
```
little-dude, line 168: Nice, we don't have many tests in `rtnetlink`, glad to see some!

little-dude (changes requested):
This is awesome, thank you for the hard and quality @wllenyj ! My only change requests are pretty minor.

netlink-packet-route/src/rtnl/tc/nlas/class/mod.rs
little-dude: I like your new implementation better, it minimizes the changes in `netlink-packet-route`. Thanks for doing this!

netlink-packet-route/src/rtnl/tc/constants.rs
wllenyj: fixed

netlink-packet-route/src/rtnl/tc/message.rs
wllenyj: fixed.

netlink-packet-route/src/rtnl/tc/nlas/action/mirred.rs
wllenyj: The first 5 fields could use the public struct TcGen, but I haven't found a good implementation.

Removed first.

netlink-packet-route/src/rtnl/tc/nlas/filter/mod.rs
wllenyj: The naming is really confusing. I tried naming the u32.rs file as U32.rs so that I could use the U32 mod. but there is a compile warning.

netlink-packet-route/src/rtnl/tc/nlas/qdisc/prio.rs
wllenyj: Remove it for now

netlink-packet-route/src/rtnl/tc/nlas/filter/mod.rs
little-dude: Yeah this is fine :)

little-dude (approved):

little-dude:
Thanks @wllenyj 

If you don't mind I'll wait for https://github.com/little-dude/netlink/pull/234 to land to make a release.

wllenyj:
> If you don't mind I'll wait for #234 to land to make a release.

That's ok.

PR #234, opened 2022-02-11
next_hops: extend to support attributes other than just RTA_GATEWAY

rshearman:
The NextHopBuffer used with RTA_MULTIPATH attributes can contain more
than just RTA_GATEWAY - it can also currently contain RTA_VIA,
RTA_FLOW, RTA_ENCAP and RTA_ENCAP_TYPE attributes, and more could be
added in the future.

So extend the NextHop struct to store a Vec<Nla> for all these other
possible attributes, reusing the route Nla type, and emit and parse
them accordingly.

This is unfortunately an API-breaking change, but is done in a way
that limit breakages as much as possible.

### DIFF
--- a/netlink-packet-route/src/lib.rs
+++ b/netlink-packet-route/src/lib.rs
@@ -23,22 +23,6 @@ pub use self::rtnl::*;
 #[macro_use]
 extern crate lazy_static;
 
-use std::net::IpAddr;
-
-pub(crate) fn ip_len(addr: &IpAddr) -> usize {
-    match addr {
-        IpAddr::V4(_) => 4,
-        IpAddr::V6(_) => 16,
-    }
-}
-
-pub(crate) fn emit_ip(buf: &mut [u8], addr: &IpAddr) {
-    match addr {
-        IpAddr::V4(ref ip) => buf.copy_from_slice(&ip.octets()),
-        IpAddr::V6(ref ip) => buf.copy_from_slice(&ip.octets()),
-    }
-}
-
 #[cfg(test)]
 #[macro_use]
 extern crate pretty_assertions;
--- a/netlink-packet-route/src/rtnl/route/nlas/next_hops.rs
+++ b/netlink-packet-route/src/rtnl/route/nlas/next_hops.rs
@@ -2,11 +2,10 @@ use anyhow::Context;
 use std::net::IpAddr;
 
 use crate::{
-    constants::{self, RTA_GATEWAY},
-    emit_ip,
-    ip_len,
-    nlas::{Nla, NlaBuffer},
+    constants,
+    nlas::{NlaBuffer, NlasIterator},
     parsers::parse_ip,
+    route::nlas::Nla,
     traits::{Emitable, Parseable},
     DecodeError,
 };
@@ -23,12 +22,14 @@ bitflags! {
     }
 }
 
+const PAYLOAD_OFFSET: usize = 8;
+
 buffer!(NextHopBuffer {
     length: (u16, 0..2),
     flags: (u8, 2),
     hops: (u8, 3),
     interface_id: (u32, 4..8),
-    gateway_nla: (slice, GATEWAY_OFFSET..),
+    payload: (slice, PAYLOAD_OFFSET..),
 });
 
 impl<T: AsRef<[u8]>> NextHopBuffer<T> {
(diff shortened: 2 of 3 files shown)


rshearman:
Pushed to fix build/test failures with rich_nlas target.

netlink-packet-route/src/rtnl/route/nlas/next_hops.rs
```diff
@@ -28,6 +27,7 @@ buffer!(NextHopBuffer {
     flags: (u8, 2),
     hops: (u8, 3),
     interface_id: (u32, 4..8),
+    // This might contain other NLAs too
     gateway_nla: (slice, GATEWAY_OFFSET..),
```
little-dude, line 31: Maybe this should be renamed `nlas` since we can have different sorts of attributes.

netlink-packet-route/src/rtnl/route/nlas/next_hops.rs
```diff
@@ -55,9 +57,15 @@ impl<T: AsRef<[u8]>> NextHopBuffer<T> {
     }
 }
 
+impl<'a, T: AsRef<[u8]> + ?Sized> NextHopBuffer<&'a T> {
+    pub fn nlas(&self) -> impl Iterator<Item = Result<NlaBuffer<&'a [u8]>, DecodeError>> {
+        NlasIterator::new(&self.gateway_nla()[..(self.length() as usize - GATEWAY_OFFSET)])
+    }
+}
+
 const GATEWAY_OFFSET: usize = 8;
```
little-dude, line 66: Not sure why I put this constant here, would you mind moving it up, above the `buffer!` call where it's first used?

netlink-packet-route/src/rtnl/route/nlas/next_hops.rs
```diff
@@ -40,8 +40,10 @@ impl<T: AsRef<[u8]>> NextHopBuffer<T> {
 
     fn check_buffer_length(&self) -> Result<(), DecodeError> {
         let len = self.buffer.as_ref().len();
-        if len < 8 {
-            return Err(format!("invalid NextHopBuffer: length {} < {}", len, 8).into());
+        if len < GATEWAY_OFFSET {
```
little-dude, line 43: Thanks

netlink-packet-route/src/rtnl/route/nlas/next_hops.rs
```diff
@@ -67,57 +75,59 @@ pub struct NextHop {
     pub interface_id: u32,
     /// Gateway address (it is actually encoded as an `RTA_GATEWAY` nla)
     pub gateway: Option<IpAddr>,
```
little-dude, line 77: What do you think of removing `gateway`? It feels weird that to be the gateway both in `self.nlas` and `self.gateway`, and I don't think there are other places where we do this.

netlink-packet-route/src/lib.rs
```diff
@@ -23,22 +23,6 @@ pub use self::rtnl::*;
 #[macro_use]
 extern crate lazy_static;
 
-use std::net::IpAddr;
```
little-dude, line 26: This is nice :)

little-dude (commented):
Sorry for the slow review. This looks good, but if you don't mind, could you make some additional cleanups since you're touching that module?

- renaming `NextHopBuffer.gateway_nla` into `nlas`
- renaming `GATEWAY_OFFSET` into `NLAS_OFFSET` and moving it to the top of the module
- removing `NextHop.gateway`

netlink-packet-route/src/rtnl/route/nlas/next_hops.rs
rshearman: Sure, done.

netlink-packet-route/src/rtnl/route/nlas/next_hops.rs
rshearman: Sure, done.

netlink-packet-route/src/rtnl/route/nlas/next_hops.rs
rshearman: Yes, that seems like a good idea. If we're making a non-backwards-compatible API change, then might as well do it properly.

little-dude:
Thanks @rshearman  :)

PR #246, opened 2022-03-06
rtnetlink: add a way to set netlink flags in requests

little-dude:
There has been several PRs to add helper methods to set the flags in
the netlink header, so it seems that this is something we should
provide systematically.

This PR introduces three new types to manipulate netlink flags:

- `NewFlags`: flags used for "set" and "add" requests
- `DelFlags`: flags used to "del" requests
- `GetFlags:` flags used for "get" requests

Each request now has a `set_flags` method to force a certain set of
flags. This may interfere with other builder methods for certain
requests, or may simply render the request ineffective, so the
documentation always warns users.

### DIFF
--- a/rtnetlink/Cargo.toml
+++ b/rtnetlink/Cargo.toml
@@ -26,6 +26,7 @@ netlink-proto = { default-features = false, version = "0.9" }
 nix = "0.22.0"
 tokio = { version = "1.0.1", features = ["rt"], optional = true}
 async-global-executor = { version = "2.0.2", optional = true }
+bitflags = "1.3.2"
 
 [dev-dependencies]
 env_logger = "0.8.2"
--- a/rtnetlink/src/addr/add.rs
+++ b/rtnetlink/src/addr/add.rs
@@ -3,27 +3,19 @@
 use futures::stream::StreamExt;
 use std::net::{IpAddr, Ipv4Addr};
 
-use netlink_packet_route::{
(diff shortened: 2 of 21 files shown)


little-dude:
@cathay4t @ffmancera I'd appreciate a review on this one if you have time, just to make sure I haven't broken anything.

wllenyj:
Will the new commit differ from the default flags of the original code? Does this mean that a major version number needs to be upgraded?

cathay4t:
New version (0.x+1) will be released as we changed removed an member of `pub struct`.

BTW, I am still reviewing the code. Personally I am reluctant on adding new dependency, need to weight the benefit and risk.

cathay4t (changes requested):
* The flags is for netlink, not rtnetlink.
 *  Differentiating AddFlag and DelFlag, add extra learning to user who already know kernel netlink. And also make the API complex.
 * Human typed integer is hard to maintain and review comparing to kernel header file is using hex format.
 *  Exposing a public struct generated by a wrapper macros is dangerous, the maintainer bitflags might hold different understanding on backwards compatibility than this project.
 * Allowing user to `set_flag` via a u16 integer is simple enough to solve this problem. Why make it complex using struct and rust macros? As you already say in comment, user who call `set_flags` should know what they are doing, those user should understand the integer use for `set_flags` is bit flags, for them an u16 integer is simpler than a struct.

Issue #44, opened 2019-04-25
Avoid defining kernel dependent data structures.


pinkisemils:
Currently, `netlink-packet::rtnl::link::inet6::dev_conf::LinkInet6DevConfBuffer` is trying to mirror the kernel's `ipv6_devconf` struct as defined in `/lib/modules/$(uname -r)/build/include/linux/ipv6.h. Since certain fields in `ipv6_devconf` are dependent on the kernel config, it'll be hard to ensure that this crate is ever 100% compatible with the machine it's running on. For instance, the `dump_links` example will not work on Debian 9. I'm unsure if there are any other structs that are kernel config dependent, but there might be.

Whilst I cannot envision nice and fluffy solution to this issue immediately, treating the device config (or any structure that can change from kernel to kernel) NLA as an opaque byte buffer would be a solution. Then a different crate can be used to try and deal with the multiplicity of different `ipv6_devconf` definitions.

For anyone looking for a patchy solution, there's this [branch](https://github.com/mullvad/netlink/tree/hack-older-kernel-compat) which hacks around the expectation of a specific length of `ipv6_devconf`, but  there might be other kernel-dependent structs.

little-dude:
Hello,

I did not even realize there were config dependent structs :(

I can't see an immediate solution either at this point, so yeah, keeping those a `Vec<u8>` might be better for the time being.

(I'll be AFK for the next 15 days, so won't be able to comment further)

squeed:
I'd been toying around with my own netlink library (though it's high time I gave it up and switched to this), and I pulled these in with `bindgen`. While binaries won't work cross-platform, it does fix local compilation...

little-dude:
FWIW it seems that the current best solution is to use different features for difference kernels: https://users.rust-lang.org/t/defining-custom-cfg-predicates/72610/4

cathay4t:
The Debian 9 is not supported any more as stated by https://www.debian.org/releases/stretch/ 
I don't see a point of support its kernel. Something not working in Debian 9 is not a good reason here.

PR #247, opened 2022-03-18
bridge: Fix decode error on bridge with IFLA_BR_MCAST_QUERIER_STATE

cathay4t:
With kernel supporting `IFLA_BR_MCAST_QUERIER_STATE`, we got error:

    Decode error occurred: Failed to parse message with type 16

After painfull debug, this is caused by IFLA_BRIDGE_FLAGS holding the
same value as `IFLA_BR_MCAST_QUERIER_STATE` and expecting the payload
been a u16.

This is incorrect:
 * `IFLA_BRIDGE_FLAGS` is 0, and for `IFLA_AF_SPEC`, not `IFLA_LINKINFO`.
 * `IFLA_BR_MCAST_QUERIER_STATE` is a nested netlink attribute, not u16
   which cause this failure.

I have no project needing the support `IFLA_BR_MCAST_QUERIER_STATE` of
it yet, so this patch just remove `IFLA_BRIDGE_FLAGS`,
`IFLA_BRIDGE_VLAN_INFO`, and their associates.

Also reordering the order of parsing `IFLA_LINKINFO` for bridge, to
match the numeric order. So the developer could easily compare it to
`/usr/include/linux/if_link.h`.

### DIFF
--- a/netlink-packet-route/src/rtnl/constants.rs
+++ b/netlink-packet-route/src/rtnl/constants.rs
@@ -395,16 +395,16 @@ pub const IFLA_INFO_XSTATS: u16 = 3;
 pub const IFLA_INFO_SLAVE_KIND: u16 = 4;
 pub const IFLA_INFO_SLAVE_DATA: u16 = 5;
 // Bridge flags
-pub const IFLA_BRIDGE_FLAGS: u16 = 47;
-pub const BRIDGE_FLAGS_MASTER: u16 = 1; /* Bridge command to/from master */
-pub const BRIDGE_FLAGS_SELF: u16 = 2; /* Bridge command to/from lowerdev */
-
-pub const IFLA_BRIDGE_VLAN_INFO: u16 = 48;
-pub const BRIDGE_VLAN_INFO_MASTER: u16 = 1;
-pub const BRIDGE_VLAN_INFO_PVID: u16 = 4;
-pub const BRIDGE_VLAN_INFO_UNTAGGED: u16 = 8;
-pub const BRIDGE_VLAN_INFO_RANGE_BEGIN: u16 = 16;
-pub const BRIDGE_VLAN_INFO_RANGE_END: u16 = 32;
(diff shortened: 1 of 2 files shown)


cathay4t:
@little-dude This bug is breaking my project. Please review and tag new release afterwards. The use of `anyhow::context` make the debug life horrible, it discard all the error in the middle, but showing top level meaningless message. Any idea how to improve it?

cathay4t:
@ffmancera The impact of this bug: it stop nmstate/nispor on all linux bridge related tasks. In case you ran into also.

little-dude:
Regarding `anyhow`: you get all the context if you print it as debug: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=832a92287ed29ef27954a25d3b32750a

I'm on vacation for two weeks so I won't be able to review this before, sorry. Is this a case where different kernels will expect a different message format as in https://github.com/little-dude/netlink/issues/44 ?

cathay4t:
No, it is not a kernel version specific difference. The last modifcation regarding `IFLA_BRIDGE_FLAGS` is:

[2469ffd72] net: set and query VEB/VEPA bridge mode via PF_BRIDGE John Fastabend (2012-10-24 04:13:03 PM)

Which is 10 years ago.

cathay4t:
Regarding the anyhow preserving lower level errors, it is not happening in this project. This is the full debug log:
https://gist.github.com/cathay4t/78d7550ae97db8254ce2891f49cda237

It only contains the error: `failed to decode packet <package_u8_array_dump_omitted>: Decode error occurred: Failed to parse message with type 16`, nothing else.

stbuehler:
It seems #222 also removes `IFLA_BRIDGE_FLAGS` and `IFLA_BRIDGE_VLAN_INFO` from `InfoBridge`, and tries to implement `IFLA_AF_SPEC` and nested key `AF_BRIDGE` with those instead.

cathay4t:
Close in favor of #222

Issue #230, opened 2022-01-24
Allow access to Ack messages containing data


mraerino:
When implementing the [W1 Netlink](https://01.org/linuxgraphics/gfx-docs/drm/w1/w1-netlink.html) adapter (uses `NETLINK_CONNECTOR`) i ran into this issue:

Responses to requests are send with the `message_type` set to `0x3`. Therefore as responses arrive, the `NetlinkPayload` enum will be set to `Done` and any payload attached to the message is thrown away.

Do you think there is a good way to model this behaviour in the library?
I think i could also make a custom codec and re-implement NetlinkMessage in a different way, but i'd rather have this solved upstream if possible.

My (very rough) code is here: https://github.com/kalkspace/w1-netlink-rs/blob/main/examples/lowlevel.rs

This is one of the payloads coming back in that example code:
```rs
[
    // netlink header
    0x34, 0x00, 0x00, 0x00, // len
    0x03, 0x00, // type: Done
    0x00, 0x00, // flags
    0x00, 0x00, 0x00, 0x00, // seq
    0x00, 0x00, 0x00, 0x00, // pid

    // netlink connector header
    0x03, 0x00, 0x00, 0x00, // idx
    0x01, 0x00, 0x00, 0x00, // val
    0x00, 0x00, 0x00, 0x00, // seq
    0x01, 0x00, 0x00, 0x00, // ack
    0x10, 0x00, 0x61, 0x38, // len (excl header) - looks wrong, idk

    // w1 message header
    0x06, // type: W1_LIST_MASTERS
    0x00, // status: no error
    0x04, 0x00, // len of data
    0x30, 0x20, 0x2D, 0x2D, 0x2D, 0x70, 0x20, 0x30, // device id
    // data
    0x01, 0x00, 0x00, 0x00, // bus id
]
```

mraerino:
@little-dude any guidance here?

little-dude:
Hello @mraerino :)

To be honest I don't know, really. I didn't expect these messages to have a payload.

## Option 1

Make the payload of Done messages a `Vec<u8>`:
```rust
#[derive(Debug, PartialEq, Eq, Clone)]
pub enum NetlinkPayload<I> {
    Done(Vec<u8>),
    Error(ErrorMessage),
    Ack(AckMessage),
    Noop,
    Overrun(Vec<u8>),
    InnerMessage(I),
}
```
But then users have to do the parsing themself. Not ideal

## Option 2

[...]

PR #250, opened 2022-04-12
Improved TokioSocket::recv() and recv_from()

mhupfer:
Pass flags to recv() and recv_from(), give back the received bytes as result.

Thats enables passing MSG_PEEK | MSG_TRUNC to determine the number of  bytes in the socket first, before allocating a buffer and receiving the data.

### DIFF
--- a/netlink-proto/src/framed.rs
+++ b/netlink-proto/src/framed.rs
@@ -61,8 +61,8 @@ where
             reader.clear();
             reader.reserve(INITIAL_READER_CAPACITY);
 
-            *in_addr = match ready!(socket.poll_recv_from(cx, reader)) {
-                Ok(addr) => addr,
+            *in_addr = match ready!(socket.poll_recv_from(cx, reader, 0)) {
+                Ok((_len, addr)) => addr,
                 Err(e) => {
                     error!("failed to read from netlink socket: {:?}", e);
                     return Poll::Ready(None);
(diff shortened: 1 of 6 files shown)


stbuehler:
`AsyncSocket::poll_recv_from_full` already provides the "get size of message first, then allocate, then read" functionality. (I just saw the doc for it is broken, and just a copy of the one above. That probably should get fixed.)

I'm quite sure I designed the trait this way on purpose, because handling flags and a length that (with the `MSG_TRUNC` flag) doesn't necessarily reflect the number of bytes read is imho quite dangerous and should be used for a low-level API only.

Any reason you can't use `AsyncSocket::poll_recv_from_full`?

mhupfer:
Found it, will give it a try.

Meanwhile I stumbled upon another issue. In impl<'buffer, B, I> Parseable<NetlinkBuffer<&'buffer B>> for NetlinkMessage<I> the code exits immediately when it detects NLMSG_DONE. But at least for the kernel connector (https://github.com/torvalds/linux/blob/master/Documentation/driver-api/connector.rst) the last messages flagged with NLMSG_DONE transports a payload.

I solved that by commenting out one line:

            ...
            //NLMSG_DONE => Done,
           ...

and checking in my application if the received buffer is completely parsed. Maybe you have a better idea ...

stbuehler:
Regarding `NLMSG_DONE` see #230 and perhaps #249.

mhupfer:
And when will #249 be merged?

mhupfer:
Can be achieved by using AsyncSocket::poll_recv_from_full

PR #251, opened 2022-04-15
Fix lint: unnecessary parentheses

stbuehler:
My editor keeps complaining about this :)


stbuehler:
Added a second commit to fix clippy lints in tests. `#[allow(clippy::unusual_byte_groupings)]` silences clippy complaints about things like `handle: 0x800_00_800` - the unusual grouping is on purpose.

@cathay4t this PR should be easy to review imho - and it would make it easier for me to maintain my branches (because they are usually based on this one, but I don't want to submit the commits here in other PRs)

JohnTitor (approved):

JohnTitor:
LGTM, merging as it's trivial. Thanks!

PR #252, opened 2022-04-23
Upgrade nix to 0.24.1, limit features

rtzoeller:
This removes an indirect dependency on memoffset, and should slightly improve compile times.


cathay4t (approved):

cathay4t:
Merged. Thanks for the contribution!

PR #253, opened 2022-04-26
Must not return/unwind from ForkResult::Child branch (neither errors nor panics)

stbuehler:
I got `Namespace error mkdir error: EACCES: Permission denied` errors in the child processes, which were returned (and panicked on further up the stack); with the (forked!) tokio multithreaded runtime this somehow didn't actually kill the child process, so it blocks forever (I think it is waiting for the "worker threads" to exit, but they don't exist in the forked process).

I think the correct solution is that you should never unwind/return once you are in the child process (from the method that ran `fork`).


cathay4t (approved):
LGTM. Thanks!

PR #259, opened 2022-05-07
Optimize codec implementations

stbuehler:
#141 mentions "have `NetlinkFramed` return a `Result` instead of an `Option`" as "todo item", and for that I had a look at the codec implementations again. It turns out `decode` actually never returns an error - and if it did, it probably should return a `DecodeError` instead of an `io::Error`.

Anyway, I tried to clean up the code a little bit so that adding proper error handling should be easier in the future.

As usual I recommend reviewing my commits individually - I split them for that purpose :)


JohnTitor (approved):
Looks good 👍

wllenyj:
Thanks for your work. 
Is it convenient to add some unit test cases?

cathay4t:
Let's move on. The codec code has been extensively used by other sub-crates, so it is tested but not explicitly.

PR #267, opened 2022-06-22
Improve internal sub-crate dependency

cathay4t:
Instead of `patch.crates-io` globally, the internal sub-crate dependency
should use `foo = { version = "0.1.0", path = "../foo" }`.

For more detail, please read
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,18 +38,3 @@ default-members = [
     "audit",
     "mptcp-pm",
 ]
-
-[patch.crates-io]
-netlink-sys = { path = "netlink-sys" }
-netlink-packet-core = { path = "netlink-packet-core" }
-netlink-packet-utils = { path = "netlink-packet-utils" }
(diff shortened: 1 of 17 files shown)


JohnTitor:
Could you elaborate on why the `path` specifier is more reasonable?

JohnTitor (approved):

stbuehler:
The patch field was created in #169 (by me), I think my specific concern was to test changes to a single crate in the repository from a different project; I think in this case cargo would still try to use the referenced path-dependencies.

I admit I didn't know cargo handles the combined `path` + `version` in a good way.

JohnTitor:
@cathay4t 
Makes sense, thanks for clarifying!

Btw, shouldn't we yank either version of netlink-packet-generic?

cathay4t:
@stbuehler In side of crate, cargo use `path`, outside uses `version`.

PR #274, opened 2022-06-29
rtnetlink: LinkDelRequest set correct flags

### LINKED ISSUE #273: rtnetlink: LinkDelRequest broken on kernel v5.19
The LinkDelRequest sets the following flags: https://github.com/little-dude/netlink/blob/15324923e655a5c786a23cbe88b5d523c46a2f3b/rtnetlink/src/link/del.rs#L39

From `netlink(7)`: NLM_F_EXCL and NLM_F_CREATE are only valid for NEW requests.
The problem is that kernel 5.19 added a new DELETE flag NLM_F_BULK which has the same value as NLM_F_EXCL, so now the delete requests fail with EOPNOTSUPP.

Here is the flag list from the kernel: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/netlink.h#n52 

To fix this we should only set NLM_F_REQUEST and NLM_F_ACK for a delete request. I will submit a PR.

Luap99:
NLM_F_EXCL and NLM_F_CREATE are not valid flags for the delete request.
On kernel v5.19 there was a new delete flag NLM_F_BULK added which has
the same value as NLM_F_EXCL. This causes the kernel to respond with
EOPNOTSUPP when trying to remove an interface.

Fixes #273

### DIFF
--- a/rtnetlink/src/link/del.rs
+++ b/rtnetlink/src/link/del.rs
@@ -3,15 +3,7 @@
 use futures::stream::StreamExt;
 
 use crate::{
-    packet::{
-        LinkMessage,
-        NetlinkMessage,
-        RtnlMessage,
-        NLM_F_ACK,
-        NLM_F_CREATE,
-        NLM_F_EXCL,
-        NLM_F_REQUEST,
-    },
+    packet::{LinkMessage, NetlinkMessage, RtnlMessage, NLM_F_ACK, NLM_F_REQUEST},
     try_nl,
     Error,
     Handle,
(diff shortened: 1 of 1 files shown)


cathay4t (approved):

cathay4t:
@Luap99  Could you run `cargo fmt` again your code?

Luap99:
done

cathay4t:
@Luap99 You will get new release before Friday(July 15).

PR #280, opened 2022-07-27
Fix panics when using iif or oif with ip rule

dxfgg17:
Copying string s (which size is `n`) in buffer of size `n+1` always panics because rust expect two slices of same length.

Since buffer has been initialized with 0, no need to add a second instruction like : `buffer[s.len()+1] = 0;`.

### DIFF
--- a/netlink-packet-route/src/rtnl/rule/nlas/mod.rs
+++ b/netlink-packet-route/src/rtnl/rule/nlas/mod.rs
@@ -116,7 +116,7 @@ impl nlas::Nla for Nla {
         }
     }
 
-    fn emit_value(&self, buffer: &mut [u8]) {
+    fn emit_value(&self, mut buffer: &mut [u8]) {
         use self::Nla::*;
         match self {
             Unspec(ref bytes)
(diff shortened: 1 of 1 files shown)


cathay4t (dismissed):

cathay4t (changes requested):
CI found this compile warning at https://github.com/little-dude/netlink/runs/7778198529?check_suite_focus=true#step:13:9

```
warning: variable does not need to be mutable
   --> netlink-packet-route/src/rtnl/rule/nlas/mod.rs:119:26
    |
119 |     fn emit_value(&self, mut buffer: &mut [u[8](https://github.com/little-dude/netlink/runs/7778198529?check_suite_focus=true#step:13:9)]) {
    |                          ----^^^^^^
    |                          |
    |                          help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default
```

dxfgg17:
> compile

Sorry for the mistake and thank you for your review. I fixed the issue :)

cathay4t (approved):

cathay4t:
Merged. Thanks!

PR #282, opened 2022-08-11
New crate `wl-nl80211` for wireless network

cathay4t:
The `nl80211` name is occupied in crates.io, so we use `wl-nl80211`
instead.

Initially we only add the support of `NL80211_CMD_GET_INTERFACE` equal
to command `iw dev`.

Except `NL80211_ATTR_MLO_LINKS` all other NLAs are tested on my laptop.

Example code is stored as `wl-nl80211/examples/dump_nl80211_iface.rs`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ members = [
     "rtnetlink",
     "audit",
     "mptcp-pm",
+    "wl-nl80211",
 ]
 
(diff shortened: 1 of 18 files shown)


cathay4t:
Closing as this is moved to https://github.com/rust-netlink/wl-nl80211/

PR #279, opened 2022-07-26
diag: add support for SOCK_DESTROY

anisse:
It is implemented on top of SockDiagMessage, and allows using Linux's
CONFIG_INET_DIAG_DESTROY feature to close an arbitrary socket.

Please consider this PR more like a request for comments, and a feature request for SOCK_DESTROY support. I'm not really a fan of how it's implemented, but I wanted to see if this was possible, and sent the first working version :-)

PR #283, opened 2022-08-11
RFC: diag: add support for SOCK_DESTROY

anisse:
Please consider this PR more like a request for comments, and a feature request for SOCK_DESTROY support. I'm not really a fan of how it's implemented, but I wanted to see if this was possible, and sent the first working version :-)

It is implemented on top of SockDiagMessage, and allows using Linux's CONFIG_INET_DIAG_DESTROY feature to close an arbitrary socket.

It was previously merged by accident in PR #279 . For an example of how to use this code, see: https://github.com/anisse/tcpkill/blob/cfd96d5dec438a3722edb25938a88aa7671fae84/src/tcpkill_netlink.rs#L21

### DIFF
--- a/netlink-packet-sock-diag/src/message.rs
+++ b/netlink-packet-sock-diag/src/message.rs
@@ -10,6 +10,7 @@ use crate::{
     NetlinkPayload,
     NetlinkSerializable,
     SockDiagBuffer,
+    SOCK_DESTROY,
     SOCK_DIAG_BY_FAMILY,
 };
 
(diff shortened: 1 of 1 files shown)


cathay4t:
@anisse We are moving splitting this project into small parts. Could you move this PR to https://github.com/rust-netlink/netlink-packet-sock-diag

For the comment part, personally,  I would like hear the use case why privileged user/process need to shutdown a arbitrary socket. I don't like dead code.

PR #290, opened 2022-08-27
Fix ip_monitor example

WGH-:
List of groups that's specified in nl_group field of the address is
actually a bitfield, whereas `RTNLGRP_*` constants are 1, 2, 3, etc,
so they must be shifted before using them in nl_group.

See iproute2 source for example[1][2].

[1] https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/ip/ipmonitor.c?id=deb485541f34cc77e5228ab6ba6e1af58014727b#n198
[2] https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/include/utils.h?id=deb485541f34cc77e5228ab6ba6e1af58014727b#n230

### DIFF
--- a/rtnetlink/examples/ip_monitor.rs
+++ b/rtnetlink/examples/ip_monitor.rs
@@ -8,6 +8,16 @@ use rtnetlink::{
     sys::{AsyncSocket, SocketAddr},
 };
 
+fn nl_mgrp(group: u32) -> u32 {
+    if group > 31 {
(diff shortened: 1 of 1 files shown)


frederictobiasc:
Hello @WGH- ! I have just stumbled across the same issue and hope your PR will get merged soon.
Thanks for the fix!

JohnTitor (approved):
LGTM, thanks!

